PR #1, opened 2015-10-01
Add missing requirement

freeman-lab:
This might be a glitch in my environment, but upon installing I found I was missing `ayncio`, and pip installing it did the trick. Not sure about the version to request, but here I just suggested the most recent one.


nchammas:
Hey @freeman-lab, I'm gonna close this PR in favor of #2, which addresses the root issue of supported Python versions and friendly error reporting.

Issue #8, opened 2015-10-05
Properly reconfigure Spark on cluster restart


nchammas:
If you stop and then start a Flintrock cluster you need to do a little work to make sure that Spark comes back up correctly.

During cluster launch Spark is installed via class methods. These methods need to be refined a bit to cover this case (and eventually, related cases like adding or removing nodes).

broxtronix:
I took a crack at this as part of #13.  It turned out that this refactor was closely related to #6, so I did them together.  The module refactor is something of a straw-man, but whatever we settle on I believe the start and stop logic now works as it should.

One thing that has not been done yet is any refinements to allow adding or removing of slave nodes on the fly.

nchammas:
I took a more limited approach to solving this in #19.

I'll briefly comment on #13 in relation to our offline conversation.

Issue #6, opened 2015-10-05
Split out provider- and service-specific code to separate Python modules


nchammas:
We should split up `flintrock.py` into a few modules that each focus on some area.

I think we should have a module for each provider (e.g. EC2, GCE) which covers all the major methods related to that provider like provisioning nodes, destroying resources, and so forth.

Additionally, we should have a Python module for each Flintrock module. A Flintrock module is a service or application you install on the cluster, like Spark. So methods that install, configure, or otherwise manage given a service like that go in one file.

broxtronix:
I took a crack an this in #13.  It's just a straw-man at this point.  Definitely eager for feedback!

PR #13, opened 2015-10-07
Improved cluster start and stop, and general refactor in preparation for GCE support.

broxtronix:
cc @freeman-lab @nchammas 

Greetings, Nick and Jeremy!   :)

This pull request is very much a work in progress, and should be thoroughly discussed and tested before we consider merging it in.

It is meant to accomplish two primary goals:

**First attempt to refactor in preparation for adding Google Compute Engine support:**

Many functions in flintrock.py have been factored out into separate files.  I chose a structure that seemed to follow the direction that things were already headed, but we should definitively discuss whether this layout will suit us well as the library grows.  At the moment, I have moved EC2 support into the flint.providers.ec2 module, and there will soon be a corresponding flint.providers.gce file.  (The top-level module name 'flint' is very much a placeholder... please help me pick a better name.)  The provider-specific functions are now part of a class, which allows us to better maintain a notion of cluster state.  That said, the choice of whether these functions should be a part of a class is certainly something we should discuss.  I also factored out all Spark installation and configuration code into flint.modules.spark.   There are also some handy utilities in flint.util, flint.ssh, and flint.async.  These encapsulate very commonly used code for running remote commands (in parallel), and these functions helped to streamline the rest of the code.

With these changes, Flintrock will be ready for the addition of GCE support.  

**Improved support for starting and stopping clusters:**

A cluster that is started after stopping now comes back up with Spark correctly configured and running.  The same configuration code that is used during cluster launch is re-used for starting the cluster.  The Spark module has been updated and improved to support this new functionality, and the install(), configure(), and configure_master() function for modules is called in the appropriate locations in launch_cluster() and start_cluster()   Launch time for cluster that are re-starting is faster than launching a cluster, as it should be.

I tested this code pretty carefully today, and I believe it has retained its previous stability and functionality.  However, I've undoubtedly missed a few things here and there, so please put it through its paces!  And since this also represents a major refactoring and restructuring of the library, let's discuss whether this is moving in the correct direction!  So, please check it out and take it out for a spin.  Let me know what you think.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,9 @@
+language: python
+python:
+  - "3.4"
+  - "3.5"
+install:
+  - "pip install -r requirements.txt"
+  - "pip install pep8"
+script:
+  - "./lint.sh"
--- a/flint/async.py
+++ b/flint/async.py
@@ -1,0 +1,14 @@
+def async_execute(async_fns):
+    # TODO: Use parameter names for run_in_executor() once Python 3.4.4 is released.
+    #       Until then, we leave them out to maintain compatibility across Python 3.4
+    #       and 3.5.
+    # See: http://stackoverflow.com/q/32873974/
+
+    import asyncio
+    loop = asyncio.get_event_loop()
+    tasks = []
+    for fn in async_fns:
+        task = loop.run_in_executor(executor=None, callback=fn)
+        tasks.append(task)
+    loop.run_until_complete(asyncio.wait(tasks))
+
--- a/flint/modules/spark.py
+++ b/flint/modules/spark.py
@@ -7,7 +7,6 @@ import paramiko
 from collections import namedtuple
 
 # Flintrock modules.
-from ..providers.ec2 import ClusterInfo
 from ..ssh import generate_ssh_key_pair
 from ..ssh import ssh_check_output
 
@@ -27,6 +26,15 @@ def get_formatted_template(path: str, mapping: dict) -> str:
 
     return formatted
 
+SparkTemplateVariables = namedtuple(
+    'SparkTemplateVariables', [
+        'name',
+        'master_host',
+        'slave_hosts',
+        'spark_scratch_dir',
+        'spark_master_opts'
+    ])
+
 # TODO: Turn this into an implementation of an abstract FlintrockModule class. (?)
 class Spark:
     def __init__(self, version):
@@ -35,20 +43,30 @@ class Spark:
     def install(
             self,
             ssh_client: paramiko.client.SSHClient,
-            cluster_info: ClusterInfo):
+            cluster_provider):
+
         """
         Downloads and installs Spark on a given node.
         """
         # TODO: Allow users to specify the Spark "distribution".
         distribution = 'hadoop1'
 
-        print("[{h}] Installing Spark...".format(
-            h=ssh_client.get_transport().getpeername()[0]))
+        host = ssh_client.get_transport().getpeername()[0]
+        print("[{h}] Installing Spark...".format(h=host))
+
+        # Collect information that is needed to fill in the templates
+        (master_instance, slave_instances) = cluster_provider.get_cluster_instances()
+        template_vars = SparkTemplateVariables(
+            name=cluster_provider.cluster_name,
+            master_host=master_instance.public_dns_name,
+            slave_hosts=[instance.public_dns_name for instance in slave_instances],
+            spark_scratch_dir='/mnt/spark',
+            spark_master_opts="")
 
         try:
             # TODO: Figure out how these non-template paths should work.
             ssh_check_output(
-                client=ssh_client,
+                ssh_client=ssh_client,
                 command="""
                     set -e
 
@@ -60,7 +78,7 @@ class Spark:
                     f=shlex.quote(
                         get_formatted_template(
                             path='./install-spark.sh',
-                            mapping=vars(cluster_info))),
+                            mapping=vars(template_vars))),
                     spark_version=shlex.quote(self.version),
                     distribution=shlex.quote(distribution)))
         except Exception as e:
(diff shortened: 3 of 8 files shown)


nchammas:
FYI: You can ignore the AppVeyor build for now. I haven't configured that correctly yet.

The Travis build is working, though.

nchammas:
Thanks for taking a crack at this @broxtronix!

Here's my initial feedback.

1. First off, I would recommend splitting up this changeset into at least 3 pull requests (probably in the following order):
  1. A PR that makes Spark work after cluster restarts.
  2. A PR that just splits up `flintrock.py` into a few modules with as little refactoring as possible.
  3. A PR that refactors stuff more aggressively (e.g. your proposed `async.py`, `ssh.py`, and `util.py` modules; the `SparkTemplateVariables` changes; etc.).

  This will make it much easier for me to review changes, and it lets smaller changes (like fixing Spark after cluster restarts) get in as quickly as possible without being blocked by more hairy stuff.

  Does that sound good to you?

  I've updated the [CONTRIBUTING guide](https://github.com/nchammas/flintrock/blob/master/CONTRIBUTING.md#capture-one-idea-in-one-pull-request) accordingly to reflect this approach to pull requests. I am carrying this over from my experience contributing to Spark where they follow this approach, but let me know if you think a different approach would work better for us here.
2. I see you repurposed the `ClusterInfo` named tuple to handle template variables. `ClusterInfo` is something I saw growing into [a class that represented a cluster](https://github.com/nchammas/flintrock/blob/c62e65042c9b8ded5743dc584ab4225d92541aca/flintrock.py#L110-L118).

  Each provider would have a method to launch a cluster, and that method would return an instance of an abstract cluster class. Every provider would offer its own implementation of that class, with a consistent API to manipulate the cluster (e.g. stop, start, destroy) or retrieve information about it (e.g. describe).

  I'm open to taking things in a different direction, but I think this will be easier to discuss when it's part of a much more limited PR.
3. > The top-level module name 'flint' is very much a placeholder... please help me pick a better name.

  Any reason we can't just call it `flintrock`? As far as I can tell, this is what most projects do (e.g. [Fabric](https://github.com/fabric/fabric/tree/master/fabric), [AsyncSSH](https://github.com/ronf/asyncssh/tree/master/asyncssh), [Flask](https://github.com/mitsuhiko/flask/tree/master/flask)).
4. > I tested this code pretty carefully today, and I believe it has retained its previous stability and functionality. However, I've undoubtedly missed a few things here and there, so please put it through its paces!

  Not sure if you noticed the script, but I use [`test-integration.sh`](https://github.com/nchammas/flintrock/blob/master/test-integration.sh) to do a comprehensive test of Flintrock's main commands when I make critical changes. Running the script will cost you money since it launches and manipulates a cluster, but it automatically covers all the major Flintrock commands.

broxtronix:
Thanks for taking the time to digest these changes, @nchammas!  Let me respond to a few of the points to move the discussion forward!

1. Although I did move the `ClusterInfo` namedtuple over into the spark module to serve template duty, the real replacement for `ClusterInfo` is actually the `AmazonEc2Provider` class in the `flintrock.providers.ec2` module.  Although this class doesn't encapsulate much functionality beyond what the free functions provided before, it's moving in the direction of the abstract base class that you are proposing.  I am also envisioning a `FlintrockProvider` base class that provides some common functionality shared between providers, and then subclasses that implement provider-specific logic for EC2, GCE, etc.    Of course, that base class does not exist yet, but I think things are now in place to make that change as a logical next step.

2. I agree that we should name the top level module `flintrock`.  The one thing we will need to do before we do that is to rename or move the 'flintrock' shell script that sits at the top level of the flintrock distribution.  How would you feel about creating a `bin/` directory and placing that script in there?  This is how I've seen python packages structured in the past.  (As an aside, I do think we should distribute flintrock as a normal python package through pipi as per your thoughts in #15... I'll share more thoughts about this over in that github issue).

3. I had not noticed the `test-integration.sh` script.  Nice!  I'll be sure to run that in the future.

Regarding the granularity of this pull request, I'll agree that this one is larger than is healthy for a mature open source software library.   I'm very much on board with the spirit and specifics of the CONTRIBUTING guide as a general practice once we really get going, and start to have users and (hopefully!) other contributors.  

However, in this early phase where we are merging together our two projects we will need to make a few comprehensive changes that cut across several issues.  I have found that this style of software development can be awkward to fully coordinate using the structures of github issues and pull requests.  I also don't want to spend too much time re-refactoring the code to make three separate pull requests, so I hope you might be willing to dive into this pull request a little deeper with me.

Of course, I realize there is a lot to digest at first glance, so I suggest we treat this pull request as principally addressing issue #6 : refactoring of library code into separate modules.  Consider the fact that this pull request addresses issue #8 as a bonus.  Not much was changed to make that work anyway.  The addition of the ssh.py, async.py, and util.py essentially wrap your existing code to make it easier to re-use.  We don't have to keep them as they are as we move forward, but I do think they are within the scope of this PR, and a net positive in terms of conciseness and readability of the rest of the code!

Assuming this is primarily addressing issue #6, I think we need to converge on these issues:

* What is the name of the top level module (seems like we've settled on `flintrock`!)

* How should the module hierarchy be structured?  Do you like the `flintrock.modules` and `flintrock.providers` breakdown?  

We could stop there if we wanted this to be an incremental step that primarily addresses the file layout of the library.  (And perhaps we should do this...)  If we wanted to delve a little deeper in this PR, I think the final question that would be most pertinent to discuss at this moment is: 

* What should the `FlintrockProvider` class (or whatever we call the new replacement to `ClusterInfo`) look like?  And how are its subclasses instantiated?  What state should in manage?  (These are basically the questions you already had in your notes about `ClusterInfo`).

Anyway, I realize this is a lot for a first "contribution" after going open source. :)  We will quickly settle down into a more conventional github open source development model very soon, I promise!  I'd love to video chat and resynchronize, and perhaps discuss some of these issues if you have time.

nchammas:
Hey @broxtronix, thanks for taking the time to talk about this PR in detail with me offline.

I know you mentioned you were planning on posting some notes here from our discussion before incorporating the changes we discussed, but I'll go ahead and post my own notes since they're still bouncing around in my head.

We agreed that the following changes looked good more-or-less as-is:

1. The fix to get Spark working after a cluster restart.
2. The high-level Python module split.

We agreed that the following high-level changes needed to be made:

1. Let's restore the isolation between SSH- and Spark-related code, and provider-specific code. Code that just needs a Linux host to work on should not have to know or query anything about any particular cloud provider. This should make it easier to reuse code and concepts when we add new providers.

  Let's also stick to IP addresses over DNS names when interacting with nodes wherever possible. It helps us avoid DNS name lookups and should make the code a bit simpler.
2. Let's restore the notion of a cluster object that represents a complete cluster configuration. This cluster object is what we want to pass around to the methods that install Spark and manipulate the cluster, and down the line this will be the object that we'll want to capture in some kind of Flintrock manifest that gets used to add nodes to an existing cluster.

  As we discussed, this will require a bunch of changes throughout the PR, especially in the `ec2.py` module. I'm not clear on how that module will look after making these changes, but my suggestion is to try stripping it down to just a collection of functions and leave the provider class abstraction for later. I don't think this should have any impact on our ability to add the GCE provider down the line.

By the way, I noticed the linter is only checking `flintrock.py` and not the new modules we're creating here. We'll need to update it at some point to make sure it covers all the code.

nchammas:
Hey @broxtronix, as we discussed offline, we're gonna put a hold on trying to refactor Flintrock to accept spark-gce. We can revisit the issue after Flintrock has developed a bunch on its own.

Thank you again for taking a crack at this and helping me understand how I want Flintrock to grow.

Issue #3, opened 2015-10-05
Setup automated testing for pull requests


nchammas:
It's hard to test Flintrock since most of what it does it orchestrate remote resources, but for starters we can run the [lint script](https://github.com/nchammas/flintrock/blob/0e8a0624bcec08cefa3b199bd136be602b4720d8/lint.sh) on all incoming pull requests.

I'll probably go with [Travis CI](https://travis-ci.org/) for Linux and OS X, and [AppVeyor](http://www.appveyor.com/) for Windows.

Automated testing status:
 - [x] Linux

  This is working after #12.
 - [ ] OS X

  Per #17, I'm gonna push this off until Travis [directly supports Python](https://github.com/travis-ci/travis-ci/issues/2312).
 - [ ] Windows

  Setting up testing for a Python project on AppVeyor is not trivial, but there are several guides to make it easier:
    * [Building Binary Wheels for Windows using Appveyor](https://packaging.python.org/en/latest/appveyor/)
    * [Python AppVeyor Demo](https://github.com/ogrisel/python-appveyor-demo)
    * [Ned Batchelder: Appveyor](http://nedbatchelder.com/blog/201509/appveyor.html)

  And here are several `appveyor.yml` files to review:
    * [Python AppVeyor Demo](https://github.com/ogrisel/python-appveyor-demo/blob/master/appveyor.yml)
    * [Hypothesis](https://github.com/DRMacIver/hypothesis/blob/master/appveyor.yml)
    * [coverage.py](https://bitbucket.org/ned/coveragepy/src/f5724e58bfed525cc830080450dc00825f003b46/appveyor.yml?at=default&fileviewer=file-view-default)

nchammas:
I'm gonna close this for now since we have a working CI build on Travis for Linux. We can reopen it when we get around to working on Windows support for Flintrock. I doubt Travis will polish up its support for OS X anytime soon.

PR #17, opened 2015-10-08
Have Travis test on both OS X and Linux

nchammas:
This should complete the Travis side of #3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+os:
+  - linux
+  - osx
 language: python
 python:
   - "3.4"


nchammas:
Ah, nevermind.

The advertisement [said](http://docs.travis-ci.com/user/multi-os/) that all we needed to do was add this to our `.travis.yml`:

```
os:
  - linux
  - osx
```

But the reality is that Travis does not currently support testing Python projects on OS X (see: https://github.com/travis-ci/travis-ci/issues/2312) without a bunch of legwork.

Since I already do my development and testing on OS X, we can do without this for now.

Issue #4, opened 2015-10-05
Add Flintrock commands for remotely executing shell commands on cluster nodes


nchammas:
Users launching clusters often need to do their own additional setup to install packages they need or the like.

We should offer one, maybe two new Flintrock commands to make it easy for users to run commands remotely on all the nodes of their cluster. I'm imagining they will work like stripped-down, Flintrock-specific versions of `pssh`.

Here's a proposal for what those commands might look like:

```
./flintrock run-command my-cluster 'yum install -y package'
./flintrock run-script my-cluster /path/to/local/script.sh
```

For starters I want to keep these commands as simple as possible. Just run the shell command or script on each of the cluster nodes in parallel and report whether they all passed or not. In the event of a failure, I might not even want to show which specific node or nodes failed, at least for an initial cut of this feature.

Anything more, like getting the output back from each of the nodes, will be pushed off to later, or redirected to utility tools like `pssh` which are better suited to the task.

nchammas:
I'm closing this issue since we have `run-command` (#26) implemented. I'm going to punt on implementing `run-script` since equivalent functionality can be composed by the user using `copy-files` (#7) and `run-command`.

PR #26, opened 2015-10-20
Add new Flintrock command: run-command

nchammas:
This partially addresses #4 by allowing you to run arbitrary shell commands against the nodes of your cluster.

Examples:

```sh
./flintrock run-command my-cluster 'touch /tmp/flintrock'
./flintrock run-command my-cluster -- yum install -y package
```

You just need to be careful to quote your command or delimit it with `--`. This prevents Flintrock from attempting to process any flags in your command, like `-y`.

### DIFF
--- a/flintrock.py
+++ b/flintrock.py
@@ -291,7 +291,7 @@ def cli(cli_context, config, provider):
     if os.path.exists(config):
         with open(config) as f:
             raw_config = yaml.safe_load(f)
-            config_map = normalize_keys(config_to_click(raw_config))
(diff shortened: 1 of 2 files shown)


nchammas:
cc @freeman-lab - I know you were interested in a feature like this. Curious to hear if you have any feedback on this first cut of a `run-command` feature from a user perspective.

nchammas:
Merging this in. @freeman-lab feel free to still chime in if you have thoughts on this feature!

PR #33, opened 2015-10-25
Fixed inability to destroy clusters in other EC2 regions

ericmjonas:
Currently we can't destroy non-us-west-1 clusters . This fixes  nchammas/flintrock/#32.


nchammas:
Good catch!

Issue #27, opened 2015-10-21
Implement a more lightweight display of launch/start/etc. progress


nchammas:
Flintrock's output is already much cleaner compared to spark-ec2's:

```
Launching 2 instances...
[52.91.67.xxx] SSH online.
[52.91.67.xxx] Installing Spark...
[52.91.213.xxx] SSH online.
[52.91.213.xxx] Installing Spark...
All 2 instances provisioned.
[52.91.67.xxx] Configuring Spark master...
Spark Health Report:
  * Master: ALIVE
  * Workers: 1
  * Cores: 1
  * Memory: 2.7 GB            
launch_ec2 finished in 0:02:52.
```

But this still needs some improvement. If you were to launch a 100-node cluster, you would get more than 200 lines of output: 100 lines showing when SSH comes online, and 100 lines for installing Spark.

This is noise.

A better user experience would be to show some kind of progress bar for each "stage" of the operation that Flintrock is executing.

So for cluster launches, we would show a single progress bar that advances as SSH becomes available across the nodes of the cluster, and another progress bar that advances as Spark is installed across the nodes of the cluster.

Maybe something like this:

```
SSH online       [####################..........]  75%
Spark installed  [######........................]  25%
```

I'll look into [Click's progress bar utility](http://click.pocoo.org/5/api/#click.progressbar), though it looks like I will probably have to find a workaround for [this issue](http://stackoverflow.com/q/32080382/877069) with having multiple progress bars being updated simultaneously.

Other considerations to keep in mind:
 - [ ] This must work well when Flintrock's output is redirected to a file.
 - [ ] We need some affordance that lets people see what node, if any, is hanging up an operation.

nchammas:
After some investigation, I think we'll have to implement our own progress indicator.

First, we don't need a bar. A counter for each stage will do:

```
Instance running   4/4
SSH online         3/4
Spark installed    1/4
```

Second, we will generally be setting progress based on some absolute measure of where we are (e.g. "2 done out of 4") instead of on a relative basis (e.g. "+1 done") because we're asking a remote resource for status instead of doing work ourselves, and that remote resource does work in unpredictable batches.

Click's progress bar does not support setting progress in this way; you must add increments of progress, which is [awkward in Flintrock](https://github.com/mitsuhiko/click/issues/447).

Issue #22, opened 2015-10-16
Upgrade to boto3


nchammas:
There is a new boto in town, [boto3](http://boto3.readthedocs.org/en/latest/).

Is it worth upgrading? Probably; it looks like the future of boto. And the earlier we do it the easier it'll probably be.

Are there any immediate benefits? How about long-term benefits?

PR #35, opened 2015-10-25
Initial spot instance support

### LINKED ISSUE #30: Support for spot instances
In the todo at the top of the flintrock script, we see:
```   
* Support for spot instances.
        - Show wait reason (capcity oversubscribed, price too low, etc.).
```

Currently it seems like a lot of the machinery for spot instances is in place, including threading the requested spot price through to `launch_ec2`. Would you accept a PR that actually made this live? the spark_ec2.py way is to launch the root node as non-spot and the workers as spot instances -- would this work for a first version?

ericmjonas:
This isn't a final pull request but more a point to begin discussion. It works successfully in `us-west-2` and `us-east-1`. 

When you make a spot instance request for N machines, this might result in up to N "reservations" for some number of machines (such that the total is the size of your spot request). Hence having to aggregate the results a bit - both types now contribute to an `all_instances` variable. 

@nchammas is this what you were thinking? Any feedback? 

Fixes #30

### DIFF
--- a/flintrock.py
+++ b/flintrock.py
@@ -513,11 +513,6 @@ def launch_ec2(
     Launch a fully functional cluster on EC2 with the specified configuration
     and installed modules.
     """
-    if spot_price:
-        print("Psych! Support for spot instances hasn't been implemented yet.", file=sys.stderr)
-        print("See: https://github.com/nchammas/flintrock/issues/30", file=sys.stderr)
-        sys.exit(1)
-
     try:
         get_cluster_instances_ec2(
             cluster_name=cluster_name,
@@ -533,31 +528,82 @@ def launch_ec2(
         vpc_id=vpc_id,
         region=region)
 
+    num_instances = num_slaves + 1
     try:
         connection = boto.ec2.connect_to_region(region_name=region)
 
-        print("Launching {c} instances...".format(c=num_slaves + 1))
+        if spot_price is not None:
+            print("Requesting {c} spot instances at a max price of {p}".format(c=num_instances, p=spot_price))
 
-        reservation = connection.run_instances(
-            image_id=ami,
-            min_count=(num_slaves + 1),
-            max_count=(num_slaves + 1),
-            key_name=key_name,
-            instance_type=instance_type,
-            placement=availability_zone,
-            security_group_ids=[sg.id for sg in security_groups],
-            subnet_id=subnet_id,
-            placement_group=placement_group,
-            tenancy=tenancy,
-            ebs_optimized=ebs_optimized,
-            instance_initiated_shutdown_behavior=instance_initiated_shutdown_behavior)
+            instance_reqs = connection.request_spot_instances(
+                price=spot_price,
+                image_id=ami,
+                count=(num_instances),
+                key_name=key_name,
+                instance_type=instance_type,
+                placement=availability_zone,
+                security_group_ids=[sg.id for sg in security_groups],
+                subnet_id=subnet_id,
+                placement_group=placement_group,
+                ebs_optimized=ebs_optimized)
 
-        time.sleep(10)  # AWS metadata eventual consistency tax.
+            pending_request_ids = [req.id for req in instance_reqs]
+
+            try:
+
+                while len(pending_request_ids) > 0:
+                    time.sleep(30)
+
+                    spot_requests = connection.get_all_spot_instance_requests(pending_request_ids)
+                    pending_request_ids = [req.id for req in spot_requests if req.state != 'active']
+
+                    if len(pending_request_ids) > 0:
+                        print("{grant} of {req} instances granted, waiting longer".format(
+                            grant=num_instances - len(pending_request_ids), req=num_instances))
+
+                print("All {num_instances} granted".format(num_instances=num_instances))
+                request_ids = [req.id for req in instance_reqs]
+
+                spot_requests = connection.get_all_spot_instance_requests(request_ids)
+                instance_ids = [sr.instance_id for sr in spot_requests]
+                reservation_resultset = connection.get_all_reservations(instance_ids)
+
+                all_instances = []
+                for reservation in reservation_resultset:
+                    all_instances += reservation.instances
+
+            except Exception as e:
+                print(e)
+                print("Canceling spot instance requests...")
+                connection.cancel_spot_instance_requests(instance_reqs)
+                # TODO add warning if any nodes actually launched instances
+
+                sys.exit(1)
+        else:
+            print("Launching {c} instances...".format(c=num_slaves + 1))
+
+            reservation = connection.run_instances(
+                image_id=ami,
+                min_count=(num_slaves + 1),
+                max_count=(num_slaves + 1),
+                key_name=key_name,
+                instance_type=instance_type,
+                placement=availability_zone,
+                security_group_ids=[sg.id for sg in security_groups],
+                subnet_id=subnet_id,
+                placement_group=placement_group,
+                tenancy=tenancy,
+                ebs_optimized=ebs_optimized,
+                instance_initiated_shutdown_behavior=instance_initiated_shutdown_behavior)
+
+            time.sleep(10)  # AWS metadata eventual consistency tax.
+
+            all_instances = reservation.instances
 
         # TODO: Move this to a reusable function and add a limit on
         #       wait time.
         while True:
-            for instance in reservation.instances:
+            for instance in all_instances:
                 if instance.state == 'running':
                     continue
                 else:
@@ -567,8 +613,8 @@ def launch_ec2(
             else:
                 break
 
-        master_instance = reservation.instances[0]
-        slave_instances = reservation.instances[1:]
+        master_instance = all_instances[0]
+        slave_instances = all_instances[1:]
 
         connection.create_tags(
             resource_ids=[master_instance.id],
@@ -593,7 +639,7 @@ def launch_ec2(
         loop = asyncio.get_event_loop()
 
         tasks = []
-        for instance in reservation.instances:
+        for instance in all_instances:
             # TODO: Use parameter names for run_in_executor() once Python 3.4.4 is released.
             #       Until then, we leave them out to maintain compatibility across Python 3.4
             #       and 3.5.
@@ -617,7 +663,7 @@ def launch_ec2(
         loop.close()
 
         print("All {c} instances provisioned.".format(
-            c=len(reservation.instances)))
+            c=len(all_instances)))
 
         master_ssh_client = get_ssh_client(
             user=user,


nchammas:
The general approach here seems fine to me. I'll go through and give specific feedback.

flintrock.py
```diff
@@ -526,42 +526,104 @@ def launch_ec2(
...
+                placement_group=placement_group,
+                ebs_optimized=ebs_optimized,
+                
+                )
+
+            spot_instance_request_ids = [sir.id for sir in instance_reqs]
+            # now we loop:
+            print("Waiting for spot instances to be granted...")
+            try:
+                while True:
```
nchammas: Is there any way to simplify this block of code? It's a bit hard to follow. Can we reuse, for example, [this pattern](https://github.com/nchammas/flintrock/blob/a1a5a7e8216b61eeaf8f33512c3da2b5142c542d/flintrock.py#L559-L568) for waiting on resources to reach a certain state?

flintrock.py
```diff
@@ -526,42 +526,104 @@ def launch_ec2(
...
+            except:
+                print("Canceling spot instance requests")
+                connection.cancel_spot_instance_requests(slave_reqs)
+                # Log a warning if any of these requests actually launched instances:
+                # (master_nodes, slave_nodes) = get_existing_cluster(
+                #     conn, opts, cluster_name, die_on_error=False)
+                # running = len(master_nodes) + len(slave_nodes)
+                # if running:
+                #     print(("WARNING: %d instances are still running" % running), file=stderr)
+                sys.exit(0)
```
nchammas: Interrupted commands should `sys.exit(1)`.

flintrock.py
```diff
@@ -526,42 +526,104 @@ def launch_ec2(
...
+                        print("All {machine_num} granted".format(machine_num = machine_num))
+                        reservation_resultset = connection.get_all_reservations(active_instance_ids)
+
+                        all_instances = []
+                        for reservation in reservation_resultset:
+                            all_instances += reservation.instances
+                            
+                        break
+                    else:
+                        print("{grant} of {req} instances granted, waiting longer".format(
```
nchammas: Would this be a good place to show the [bid status](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-bid-status.html)? We don't have to do it in this PR, but I just want to mention it.

I've been bitten in the past while using spark-ec2 by waiting unknowingly for requests that typically don't get fulfilled, e.g. due to `capacity-oversubscribed` or `price-too-low`.

flintrock.py
```diff
@@ -526,42 +526,104 @@ def launch_ec2(
         vpc_id=vpc_id,
         region=region)
 
+    machine_num = num_slaves + 1
```
nchammas: Nitpick: For consistency, we should probably call this `num_instances`.

flintrock.py
```diff
@@ -526,42 +526,104 @@ def launch_ec2(
...
-
-        time.sleep(10)  # AWS metadata eventual consistency tax.
-
-        # TODO: Move this to a reusable function and add a limit on
-        #       wait time.
-        while True:
-            for instance in reservation.instances:
-                if instance.state == 'running':
-                    continue
+        #spot instance request
```
nchammas: Nitpick: I'd generally prefer not to have comments explaining something if it's straightforward.

flintrock.py
```diff
@@ -573,7 +635,7 @@ def launch_ec2(
             tags={
                 'flintrock-role': 'slave',
                 'Name': '{c}-slave'.format(c=cluster_name)})
-
+
```
nchammas: Style: Blank lines should not have any whitespace.

Travis (which currently just runs `lint.sh`) will flag problems like these for you, so they should be easy to find and fix. You can see this in the PR build status.

nchammas:
Thanks for taking a crack at this! I left some specific feedback. My main concern is about the complexity of the main loop to wait for the spot instances. Otherwise, I think we can get this in for Flintrock's initial release.

By the way, you need to pull the latest changes into your fork. You're missing some [changes I introduced yesterday](https://github.com/nchammas/flintrock/blob/a1a5a7e8216b61eeaf8f33512c3da2b5142c542d/flintrock.py#L516-L519) which you'll need to remove.

flintrock.py
nchammas: On second thought, perhaps for starters we should just wait for the spot instances and not print anything here. The repeated "waiting longer" messages could be spammy.

What's your feeling on this?

flintrock.py
ericmjonas: I've cleaned up the code considerably in the request code. The challenge is we have a 1 spot request -> many reservations model, and so it's just a bit tricky. See the latest commits to this branch.

flintrock.py
ericmjonas: fixed

flintrock.py
ericmjonas: I have found that it's nice to have this heartbeat -- sometimes it can take up to 20 min for a large spot instance to be granted, even if you're far above the current market price.

flintrock.py
ericmjonas: fixed

flintrock.py
ericmjonas: fixed

flintrock.py
ericmjonas: Fixed, new pr should work

flintrock.py
```diff
@@ -526,42 +526,98 @@ def launch_ec2(
...
+                instance_type=instance_type,
+                placement=availability_zone,
+                security_group_ids=[sg.id for sg in security_groups],
+                subnet_id=subnet_id,
+                placement_group=placement_group,
+                ebs_optimized=ebs_optimized)
+
+            spot_instance_request_ids = [sir.id for sir in instance_reqs]
+
+            print("Waiting for spot instances to be granted...")
```
nchammas: Do we need this second print here if we've already told the user we've requested spot instances? Maybe we should just add some ellipsis to the print in [L534](https://github.com/nchammas/flintrock/pull/35/files#diff-1ec778e3c6cde832d96a7c08ba31f98eR534) to suggest we are waiting and get rid of this print here.

flintrock.py
```diff
@@ -526,42 +526,98 @@ def launch_ec2(
...
+        if spot_price is not None:
+            print("Requesting {c} spot instances at a max price of {p}".format(c=num_instances, p=spot_price))
+
+            instance_reqs = connection.request_spot_instances(
+                price=spot_price,
+                image_id=ami,
+                count=(num_instances),
+                key_name=key_name,
+                instance_type=instance_type,
+                placement=availability_zone,
```
nchammas: I noticed in an earlier version of this PR you were experimenting with the `launch_group` parameter. Can you tell me a bit about that? According to [the docs](http://boto.readthedocs.org/en/latest/ref/ec2.html#boto.ec2.connection.EC2Connection.request_spot_instances), it sounds useful here, but I'm not sure.

flintrock.py
```diff
@@ -526,42 +526,98 @@ def launch_ec2(
...
+            print("Waiting for spot instances to be granted...")
+            try:
+                while True:
+                    time.sleep(30)
+
+                    # Note that there can be outstanding spot instance requests
+                    # that are not part of our current requests (say, from
+                    # another process), and so we have to filter for just
+                    # our requests
+                    all_account_reqs = connection.get_all_spot_instance_requests()
```
nchammas: I think we can avoid some of the filter logic below if we only request the pending spot requests here by passing in [`request_ids`](http://boto.readthedocs.org/en/latest/ref/ec2.html#boto.ec2.connection.EC2Connection.get_all_spot_instance_requests). Here's [an example](http://www.imlucas.com/post/55003108849/waiting-for-spot-instances-to-be-fulfilled-with) of this.

flintrock.py
ericmjonas: Great idea, will add

flintrock.py
```diff
@@ -526,42 +526,98 @@ def launch_ec2(
...
+                placement_group=placement_group,
+                tenancy=tenancy,
+                ebs_optimized=ebs_optimized,
+                instance_initiated_shutdown_behavior=instance_initiated_shutdown_behavior)
+
+            time.sleep(10)  # AWS metadata eventual consistency tax.
+
+            # TODO: Move this to a reusable function and add a limit on
+            #       wait time.
+            while True:
```
nchammas: Do we not need this block to wait for the instances to come up if we are requesting spot instances?

flintrock.py
ericmjonas: We basically never use launch groups -- I was just following what spark_ec2 was doing but then cleaned up the parts that didn't seem mission-critical at the time. There are a lot of situations where having a few of your spot instances die is ok for a running spark job.

flintrock.py
ericmjonas: Great find!

flintrock.py
ericmjonas: I think we already do the appropriate (consistency) wait in the spot-instance request code. After our initial request, we wait 30s for the spot instances to be fulfilled.

nchammas: I mean the wait for the instances to get to a `running` state. We do that so the instances are guaranteed to have been assigned IP addresses, which the rest of the launch code will expect.

Right now we don't explicitly wait for the instances be `running`, so I'm wondering if we should move this block of code outside the `else` clause. Does that make sense?

nchammas:
Thanks for these latest changes @ericmjonas! The PR looks much better. I went through again and left more comments, which you are already responding to (sweet!). Basically, I still think we can greatly simplify the logic for waiting on the spot requests, but otherwise this is shaping up nicely.

A few notes to myself for the future:
* When we upgrade to boto3 (#22) we can probably replace the custom spot request wait logic entirely with [`EC2.Waiter.SpotInstanceRequestFulfilled`](http://boto3.readthedocs.org/en/latest/reference/services/ec2.html#EC2.Waiter.SpotInstanceRequestFulfilled).
* It would be nice in the future to replace the repeated print statements in the spot request wait loop with a progress bar (like proposed in #27) or a simple spinner.

flintrock.py
nchammas: Okie doke, makes sense.

ericmjonas:
@nchammas I think I messed up on my attempt to rebase to your master and now for some reason my tree is really ugly. I could try fixing but I'm guessing that we'll just squash the entire commit hist for the eventual final PR?

nchammas:
Yeah, before merging the PR I'll ask you to squash it into one commit. Don't worry about it for now.

By the way, the way I've pulled in changes to personal forks in the past is simple. If you cloned your fork locally using GitHub, you should have a remote already setup for you called `nchammas`.

```sh
# from your fork, on master
git remote -v
git pull nchammas master
# switch to your development branch, then
git rebase master
```

flintrock.py
```diff
@@ -533,31 +528,82 @@ def launch_ec2(
...
+                security_group_ids=[sg.id for sg in security_groups],
+                subnet_id=subnet_id,
+                placement_group=placement_group,
+                ebs_optimized=ebs_optimized)
+
+            spot_instance_request_ids = [sir.id for sir in instance_reqs]
+
+            try:
+                while True:
+                    time.sleep(30)
```
nchammas: I haven't launched spot instances in a while. Is this typically how long we want to wait between requests? Seems a bit long, maybe it's appropriate.

flintrock.py
```diff
@@ -533,31 +528,82 @@ def launch_ec2(
...
+                        reservation_resultset = connection.get_all_reservations(active_instance_ids)
+
+                        all_instances = []
+                        for reservation in reservation_resultset:
+                            all_instances += reservation.instances
+                        break
+                    else:
+                        print("{grant} of {req} instances granted, waiting longer".format(
+                            grant=len(active_instance_ids), req=num_instances))
+            except:
```
nchammas: We shouldn't have bare excepts like this.

```python
except Exception as e:
    print(e)
    print("Canceling spot instance requests...")  # added ellipsis here
    ...
```

This is a bit more user-friendly.

flintrock.py
```diff
@@ -533,31 +528,82 @@ def launch_ec2(
...
+                print("Canceling spot instance requests")
+                connection.cancel_spot_instance_requests(instance_reqs)
+                # FIXME add warning if any nodes actually launched instances
```
nchammas: Nit: I've been using `TODO: ...` for notes like this.

flintrock.py
```diff
@@ -533,31 +528,82 @@ def launch_ec2(
...
+                placement_group=placement_group,
+                ebs_optimized=ebs_optimized)
+
+            spot_instance_request_ids = [sir.id for sir in instance_reqs]
+
+            try:
+                while True:
+                    time.sleep(30)
+
+                    reqs = connection.get_all_spot_instance_requests(spot_instance_request_ids)
```
nchammas: This is better, but I was hoping we could really simplify things, like:

```python
pending_request_ids = [req.id for req in instance_reqs]

while pending_request_ids:
    spot_requests = connection.get_all_spot_instance_requests(pending_request_ids)
    pending_request_ids = [req.id for req in spot_requests if req.state != 'active']
```

What do you think?

nchammas:
Whoops! Sorry Eric, I missed the fact that you submitted some updates yesterday. I just went through and made another round of comments.

flintrock.py
ericmjonas: I think so -- even small spot instance requests (say for 5 instances) can take ~90 seconds to be approved (by whatever method Amazon uses to make sure you're not going to break everything) and then 2-3 min to grant, depending on price.

nchammas:
I'm having a little trouble applying this patch locally so I can test it out. Is it working for you in this state?

flintrock.py
```diff
@@ -533,31 +528,80 @@ def launch_ec2(
...
+                    spot_requests = connection.get_all_spot_instance_requests(pending_request_ids)
+                    pending_request_ids = [req.id for req in spot_requests if req.state != 'active']
 
-        time.sleep(10)  # AWS metadata eventual consistency tax.
+                    if len(pending_request_ids) > 0:
+                        print("{grant} of {req} instances granted, waiting longer".format(
+                            grant=num_instances - len(pending_request_ids), req=num_instances))
+
+                print("All {num_instances} granted".format(num_instances=num_instances))
+                spot_requests = connection.get_all_spot_instance_requests(pending_request_ids)
```
nchammas: If I'm not mistaken, `pending_request_ids` will be empty here because they are all fulfilled, so this won't work. You probably want to reuse `instance_reqs` from earlier, no?

ericmjonas: I'm sorry, I'm not sure how that made it through; fixing now.

nchammas:
Can you add "Fixes #30." to the PR description, by the way? That way when we merge this PR in GitHub will automatically close the issue for us.

nchammas:
@ericmjonas, if you're ready, I think now's a good time to squash these commits into one. If you need help doing that, pop in to the [Flintrock Gitter room](https://gitter.im/nchammas/flintrock) and I'll help you out. Once we have that done I think I can merge this in!

flintrock.py
```diff
@@ -533,31 +528,82 @@ def launch_ec2(
...
+                reservation_resultset = connection.get_all_reservations(instance_ids)
+
+                all_instances = []
+                for reservation in reservation_resultset:
+                    all_instances += reservation.instances
+
+            except Exception as e:
+                print(e)
+                print("Canceling spot instance requests...")
+                connection.cancel_spot_instance_requests(instance_reqs)
```
nchammas: Here you want `pending_request_ids`.

nchammas:
Alright, I tested this out and it looks good to me. There's an issue with the exception handling, but that's OK. I'll pull your changes in to a new PR and clean that up along with some minor style issues and get this merged in.

Thanks for working on this!

Issue #28, opened 2015-10-21
Improve root volume setup and add some EBS volume support


nchammas:
There are probably a few things that need to be done here which need to be fleshed out, including:
- [ ] Tuning the size or type of the root volume (magnetic vs. SSD; ext4 setup; etc.)
- [ ] Configuring additional EBS volumes (persistent storage)
- [ ] Configuring additional instance store volumes (ephemeral storage); good as scratch space
- [ ] Picking good defaults for this stuff
- [ ] Deciding on how much of it should be configurable

broxtronix:
I have been needing a larger root drive on my instances (10GB is not nearly enough!!), and have added the capability to choose the EBS size and type for the root drive.  As a nice fringe benefit, EBS volumes backed by SSD drives (type "gp2") are much more performant than the magnetic drive-backed EBS volumes.  With an SSD root drive, instances boot and install software much quicker.  These SSD-backed EBS volumes only cost twice as much as magnetic-backed EBS, which is well worth the cost in my opinion.  I think a 20 GB gp2 drive would be a sensible default for the root partition.

Here is the code snippet that creates a root drive (of a certain size and type).  On my Amazon Linux AMI, the root drive was automatically resized to take advantage of all of the available space allotted to it on the underlying block store.  I suspect this happens on CentOS AMIs as well.

```
           # Configure instance root drive. Note that if you choose a size
           # larger than 10GB, you may need to resize the filesystem once your
           # instance is running, using resize2fs, although some AMIs do this automatically.
           from boto.ec2.blockdevicemapping import BlockDeviceMapping, BlockDeviceType, EBSBlockDeviceType
           block_map = BlockDeviceMapping()
           device = EBSBlockDeviceType()
           device.size = int(ebs_root_vol_size)   # size in GB
           device.volume_type = ebs_root_vol_type # [ standard | gp2 | io1 ]
           device.delete_on_termination = True
           block_map["/dev/xvda"] = device

           reservation = connection.run_instances(
               image_id=ami,
               min_count=(num_slaves + 1),
               max_count=(num_slaves + 1),
               key_name=key_name,
               instance_type=instance_type,
               placement=availability_zone,
               security_group_ids=[sg.id for sg in security_groups],
               subnet_id=subnet_id,
               placement_group=placement_group,
               tenancy=tenancy,
               block_device_map=block_map,
               ebs_optimized=ebs_optimized,
               instance_initiated_shutdown_behavior=instance_initiated_shutdown_behavior)
[...]

PR #41, opened 2015-11-03
Add HDFS module

### LINKED ISSUE #20: Allow for clusters to be launched without Spark installed
Perhaps this is not a common use case, but Flintrock's internals should be cleaned up to properly support launching a cluster that does not have Spark installed.

Currently, there is some ambiguity over how the `--no-install-spark` option should work. Does it kick in automatically if the user's `config.yaml` file has no Spark module specified?

### LINKED ISSUE #40: Add support for HDFS module
HDFS is a common part of people's "Big Data" workflows. Typically, their data lives there so that's where their work starts from.

In Flintrock's case, people are more likely to want to use HDFS as a temporary store for data pulled in from other places like S3 to speed up access to it by Spark. spark-ec2 directly supports this use case with both ephemeral and permanent HDFS installations provided on launched clusters.

Flintrock should offer something similar.

nchammas:
I have HDFS working, which you can test by running `hadoop/bin/hdfs`. Spark can also read stuff from HDFS with a simple `sc.textFile('hdfs:///file')`. And clusters can now be launched with any combination of HDFS or Spark installed by specifying `--no-install-hdfs` or `--no-install-spark`, or by setting the equivalent settings in `config.yaml`.

TODOs:
 - [x] Get Spark to talk to HDFS.
 - [x] Enforce Spark config dependency on HDFS install if necessary.
    * e.g. Add HDFS options to `ClusterInfo`.
 - [x] Clean up CLI option handling.
    * e.g. `spark_version` is required if `install_spark` is set.
 - [x] Allow neither HDFS nor Spark to be installed, per #20.
 - [x] Confirm HDFS comes up correctly after cluster restart.
 - [x] Abstract out login user from Hadoop config files.
 - [x] Clean up template handling for `slave_hosts`.

Fixes #40.
Fixes #20.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ Here's a quick way to launch a cluster on EC2, assuming you already have an [AWS
 ```sh
 flintrock launch test-cluster \
     --num-slaves 1 \
+    --spark-version 1.5.1 \
     --ec2-key-name key_name \
     --ec2-identity-file /path/to/key.pem \
     --ec2-ami ami-146e2a7c \
@@ -141,6 +142,10 @@ Flintrock lets you persist your desired configuration to a file (called `config.
 
 ```yaml
 provider: ec2
+
+modules:
+  spark:
+    version: 1.5.1
 
 launch:
   num-slaves: 1
--- a/config.yaml.template
+++ b/config.yaml.template
@@ -13,10 +13,10 @@ providers:
     instance-type: m3.medium  # must be 64-bit; small instances won't work
     region: us-east-1
     # availability-zone: <name>
-    # ami: ami-146e2a7c   # Amazon Linux
-    # user: ec2-user
-    ami: ami-61bbf104   # CentOS 7
-    user: centos
+    ami: ami-60b6c60a   # Amazon Linux
+    user: ec2-user
+    # ami: ami-61bbf104   # CentOS 7
+    # user: centos
     # spot-price: <price>
     # vpc-id: <id>
     # subnet-id: <id>
--- a/config.yaml.template
+++ b/config.yaml.template
@@ -1,6 +1,8 @@
 modules:
   spark:
     version: 1.5.1
+  hdfs:
+    version: 2.7.1
 
 provider: ec2
 
@@ -25,3 +27,5 @@ providers:
 
 launch:
   num-slaves: 1
+  # install-hdfs: False
+  # install-spark: False
--- a/flintrock.py
+++ b/flintrock.py
@@ -110,11 +110,28 @@ ClusterInfo = namedtuple(
     'ClusterInfo', [
         'name',
         'ssh_key_pair',
+        'user',
         'master_host',
         'slave_hosts',
         'spark_scratch_dir',
         'spark_master_opts'
     ])
+
+
+def cluster_info_to_template_mapping(cluster_info: ClusterInfo) -> dict:
+    """
+    Convert a ClusterInfo tuple to a dictionary that we can use to fill in template
+    parameters.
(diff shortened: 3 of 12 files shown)


nchammas:
@shivaram - Any quick pointers on what to do to get Spark talking to HDFS via `sc.textFile()`?

I know HDFS is working because I can `dfs -put` and `dfs -cat` files from the master or slaves. But something is missing to allow a command like `sc.textFile('hdfs:///file.txt').count()` to work. Even `sc.textFile('hdfs://master_ip:9000/file.txt')` doesn't seem to work.

I looked through Spark's config guide and tried adding more configs like `HDFS_URL`, but I'm just bumbling about and none of it seems to work.

broxtronix:
@nchammas, I think I had basically gotten this working in spark-gce, although I never tested it much beyond the simple loading of data, and at one point it stopped working so this may not be the complete solution.  Beware!   But in case it is helpful, [here](https://github.com/broxtronix/spark-gce/tree/master/spark_gce/support_files/templates/spark/conf) are my spark config files, which include a few Hadoop-specific configurations.  In particular, take a look at `core-site.xml` (which would be placed in the `spark/conf` dir, or perhaps use the Hadoop version of the file and include it in the Spark classpath... see docs about this [here](https://spark.apache.org/docs/0.9.1/hadoop-third-party-distributions.html)).  For setting up classpaths, `SPARK_SUBMIT_LIBRARY_PATH`, `SPARK_SUBMIT_CLASSPATH`, and `HADOOP_HOME` environment vars in `spark-env.sh` seemed to do the trick.  I think the `core-site.xml` and classpath settings are the most important, as described [here].  Hope this helps!

nchammas:
@broxtronix - Oh, good point about the class path. I think I missed that.

From ["Inheriting Cluster Configuration"](https://spark.apache.org/docs/latest/hadoop-third-party-distributions.html#inheriting-cluster-configuration):

> ## Inheriting Cluster Configuration
> If you plan to read and write from HDFS using Spark, there are two Hadoop configuration files that should be included on Spark’s classpath:
>
> * `hdfs-site.xml`, which provides default behaviors for the HDFS client.
> * `core-site.xml`, which sets the default filesystem name.
> 
> The location of these configuration files varies across CDH and HDP versions, but a common location is inside of `/etc/hadoop/conf`. Some tools, such as Cloudera Manager, create configurations on-the-fly, but offer a mechanisms to download copies of them.
>
> To make these files visible to Spark, set `HADOOP_CONF_DIR` in `$SPARK_HOME/spark-env.sh` to a location containing the configuration files.

shivaram:
Yeah the core-site.xml needs to be in the classpath. In spark-ec2 we just copy it to /root/spark/conf. BTW another way to do this is to open the HDFS ui at <host>:50070 and that has the HDFS url written on it. It might be 8020 instead of 9000 on some newer versions I guess (I saw https://ambari.apache.org/1.2.3/installing-hadoop-using-ambari/content/reference_chap2_1.html)

nchammas:
OK, this is in a good state now. I'm going to polish it up a bit and merge it in the next few days if there is no feedback. The next key issue to work on will probably be #28, which we'll need if we want HDFS deployed on ephemeral storage.

One thing I'm particularly happy with in this PR is how little configuration was needed to get HDFS working. HDFS has changed a lot over the years, and it looks like spark-ec2 built up a lot of configuration "weight" supporting that technical churn. Since we only support the latest and greatest, this is a burden Flintrock doesn't have to bear. Of course, we can add more configuration as we need to, but it's good that we're starting light.

PR #51, opened 2015-11-12
[WIP] Configure ephemeral storage

nchammas:
This is a work in progress.

If the instance [offers them](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/InstanceStorage.html#instance-store-volumes), Flintrock now automatically configures up to 12 ephemeral ext4-formatted volumes. The volumes are formatted in parallel, so the time it takes to format them is dependent only on the size of the largest volume, not on the number of volumes.

This means, for example, that Flintrock can setup 12 ephemeral volumes, 2TB each, in roughly 4 minutes (e.g. d2.4xlarge instances). I think that's pretty good.

I'm not happy with adding so much involved Bash, but I wonder if there is a cleaner way to accomplish what we want here. Perhaps later on we can convert this to Python.

It would be good to see what others think of this patch, and perhaps get it tested against a wider set of instance types than I've tested against. I've tested this against t2.small, m3.medium, d2.xlarge, d2.2xlarge, and d2.4xlarge instance types with Amazon Linux.

I couldn't get this working yet with CentOS because the base AMI they offer makes it very difficult to install GNU parallel from source: no wget, no bzip2, and no perl. There's also no official yum repository offering GNU parallel to make this easier. So maybe I'll have to convert some of this stuff to Python sooner than planned.

We will still need to configure HDFS to use these ephemeral volumes. That will come in a separate PR.

### DIFF
--- a/flintrock.py
+++ b/flintrock.py
@@ -34,6 +34,7 @@ Distant future:
 import os
 import posixpath
 import errno
+import string
 import sys
 import shlex
 import subprocess
@@ -647,6 +648,12 @@ def get_ec2_block_device_map(
     block_device_map = boto.ec2.blockdevicemapping.BlockDeviceMapping()
     block_device_map[image.root_device_name] = root_device
 
+    for i in range(12):
+        ephemeral_device = boto.ec2.blockdevicemapping.BlockDeviceType(
+            ephemeral_name='ephemeral' + str(i))
+        ephemeral_device_name = '/dev/sd' + string.ascii_lowercase[i + 1]
+        block_device_map[ephemeral_device_name] = ephemeral_device
+
     return block_device_map
 
 
@@ -955,6 +962,21 @@ def provision_node(
             """.format(
                 private_key=shlex.quote(cluster_info.ssh_key_pair.private),
                 public_key=shlex.quote(cluster_info.ssh_key_pair.public)))
+
+        with client.open_sftp() as sftp:
+            sftp.put(
+                localpath='./setup-ephemeral-storage.sh',
+                remotepath='/tmp/setup-ephemeral-storage.sh')
+            sftp.chmod(path='/tmp/setup-ephemeral-storage.sh', mode=0o755)
+
+        print("[{h}] Configuring ephemeral storage...".format(h=host))
+        ssh_check_output(
+            client=client,
+            command="""
+                set -e
+                /tmp/setup-ephemeral-storage.sh
+                rm -f /tmp/setup-ephemeral-storage.sh
+            """)
 
         # The default CentOS AMIs on EC2 don't come with Java installed.
         java_home = ssh_check_output(
--- a/setup-ephemeral-storage.sh
+++ b/setup-ephemeral-storage.sh
@@ -1,0 +1,51 @@
+#!/bin/bash
+set -e
+
+# Install GNU parallel.
+if [ ! $(command -v parallel) ]; then
+    pushd /tmp
+    PARALLEL_VERSION="20151022"
+    wget "http://ftpmirror.gnu.org/parallel/parallel-${PARALLEL_VERSION}.tar.bz2"
+    bzip2 -dc "parallel-${PARALLEL_VERSION}.tar.bz2" | tar xvf -
+    pushd "parallel-${PARALLEL_VERSION}"
+    ./configure --prefix=/usr  # Amazon Linux root user doesn't have /usr/local on its $PATH
+    make
+    sudo make install
+    popd
+    rm -rf "./parallel-${PARALLEL_VERSION}*"
+    popd
+
+    # Suppress citation notice.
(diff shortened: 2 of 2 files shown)


nchammas:
Alright, I've converted this ephemeral storage setup script to Python. This should make it easier to maintain; it also makes it work on CentOS.

nchammas:
Some quick questions for people who may be interested in this PR (e.g. @ericmjonas, @shivaram, @broxtronix):
* Is the logic correct? Basically, format and mount every non-root device as an ext4 filesystem and mount it under `/mntN`.
* Should the mount points be owned by the default user (e.g. `ec2-user`) instead of root?
* Does this patch work for you?
* Any other thoughts?

shivaram:
Formatting and mounting the disks to `/mntN` is good. One more thing is that the HDFS needs to have a directory in each disk (say `/mntN/hdfs`) and these dirs need to be added to the HDFS conf.

Having the mountpoints owned by `ec2-user` seems like a good idea if Spark, HDFS will run as `ec2-user`.

nchammas:
OK, I've updated things so that the non-root user owns the mounted paths.

Additionally, I've reorganized where the devices get mounted to with the intention that, eventually, they'll look like this:

```
/media
    /root
    /ephemeral[0-N]
    /persistent[0-N]
```

Key:
* `root`: Just a regular folder that maps to the instance's root device.
* `ephemeral[0-N]`: Maps to instance store volumes.
* `persistent[0-N]`: Maps to EBS volumes.

I've put this stuff under `/media` to follow the pattern Amazon Linux seems to follow. Right now only the ephemeral volumes are configured; I'll configure the rest in separate PRs.

When we configure HDFS and Spark to use these volumes, we'll need to add a way to communicate back to Flintrock how many volumes were discovered and mounted so we can set those in the appropriate configuration files, as @shivaram pointed out.

nchammas:
I'll leave this open for a few days in case anyone has any other comments, then I'll merge it in. The next thing to work on will be configuring HDFS and Spark to use this ephemeral storage.

engrean:
This feature is very important to us as well. We use m3.2xlarge's a lot and we need all of 160 GB (2x80 GB SSD) we can get. Basically, we have to keep using EMR until we can use all of SSD's.

broxtronix:
This patch is looking great to me.  The logic seems to correctly deal with the inconsistent and somewhat haphazard mounting and formatting defaults on the Amazon and CentOS images.  Formatting in parallel is nice and fast, and I agree a Python script is going to ultimately be more maintainable and portable than a bash script.  I prefer them, anyway.

As far as ownership of these mount points is concerned, I think you've made the right choice going with `ec2-user`.  However, just FYI, things over on GCS may make this a shade more complicated.  By default on all of Google's instances you are logged into a user account matching you OAuth2 credentials.  For example, when I ssh into my GCS nodes, I log in as user `broxton`, my home directory is `/home/broxton` and the account is linked to the OAuth2 credential and permissions associated with my Google account.  If you (as another user of the same cluster) were to log in, you'd be given a separate account, home directory, etc. associated with your own credentials.  In essence you get something very much like the old UNIX NIS service, but configured for free when you start GCS instances and tied in with all of the access controls on the Google cloud platform.  (This is very convenient in a lot of ways, and one thing I like a lot better about using GCS that AWS.) 

However, there is not (by default) a "default" user account like `ec2-user` on GCS instances.  In a one user per cluster situation, this is no big deal because you can simply install Spark, Hadoop, etc. on your own OAuth2 credentialed user account and treat your own account as the de facto default user.   However, if you want multiple users to be able to log into the same cluster and share cluster resources, it becomes necessary to think more carefully about users, groups, permissions, ownership of filesystems and authentication to network services.  

In [spark-gce](https://github.com/broxtronix/spark-gce) I sidestepped these issues and settled on a one user per cluster approach by simply installing everything under the OAuth2 account that first launched the cluster.  I think this is a fine compromise, as it greatly simplifies things and does not rule out too many of the most common use cases (that I had, anyway).   However, it does mean that I cannot set up a cluster for my colleague and then ask him to start it up and use it, or share a cluster, among other things.  

I suspect Flintrock should adopt a similar one user per cluster policy and just stick with the path of least resistance which is to install everything under the `ec2-user`'s account and configure permissions accordingly.  I just wanted to bring this up as a thing worth thinking about, especially as you work towards supporting GCE.

This is a great patch, though, @nchammas.  Definitely adding a very necessary piece of cluster configuration!

nchammas:
@broxtronix - Thanks for the excellent explanation of the setup over on GCE. Yeah, I agree it will probably be best when we get there to just have Flintrock install stuff under the user logging in, which it detects via `$(logname)`: `ec2-user` on Amazon Linux; `centos` on CentOS.

Thanks also for testing this patch out! It really helps me move forward with confidence when I know others are poking around and testing out stuff. It's invaluable.

PR #59, opened 2015-11-19
Configure HDFS and Spark to use available ephemeral storage

### LINKED ISSUE #28: Improve root volume setup and add some EBS volume support

nchammas:
This change configures HDFS and Spark to use ephemeral storage when available.

The logic is as follows:
* If the cluster offers no ephemeral storage, point HDFS and Spark to the root volume, which is typically EBS-backed.
* If the cluster offers ephemeral storage, point HDFS and Spark at that and do not use the root volume.
  * In some cases, the available ephemeral storage will be smaller than the root volume, which may be confusing.

    m3.medium instances, for example, offer only 4GB of ephemeral storage, whereas the root volume is 30GB. That means you'll have *less* space available to HDFS and Spark with m3.medium instances than with, say, t2.small instances, since the latter have no ephemeral storage.

    In other words, HDFS gets 4GB of ephemeral storage with m3.medium instances, but 30GB of EBS storage from the root volume with t2.small instances.
* In no cases will HDFS or Spark be configured to use both the root volume and ephemeral volumes, since we would be mixing persistent with ephemeral storage. If this can be shown to be safe, I'm willing to change things so that both are used simultaneously.

Some of Flintrock's internals are getting a bit messier with this patch, but they will be relatively straightforward to refactor in the future. My focus for now is to get the primary functionality working.

I think we should discuss adding support for additional EBS volumes in a separate issue, so this PR fixes #28.

### DIFF
--- a/templates/hadoop/conf/hdfs-site.xml
+++ b/templates/hadoop/conf/hdfs-site.xml
@@ -9,6 +9,6 @@
 
   <property>
     <name>dfs.datanode.data.dir</name>
-    <value>/mnt/hdfs</value>
+    <value>{root_ephemeral_dirs}</value>
   </property>
 </configuration>
--- a/README.md
+++ b/README.md
@@ -198,9 +198,11 @@ Flintrock is really fast. This is how quickly it can launch fully operational cl
 
 The spark-ec2 launch times are sourced from [SPARK-5189](https://issues.apache.org/jira/browse/SPARK-5189).
 
-### Low-level Provider Options
+### Advanced Storage Setup
 
-#### EC2
+Flintrock automatically configures any available [ephemeral storage](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/InstanceStorage.html) on the cluster and makes it available to installed modules like HDFS and Spark. This storage is fast and is perfect for use as a temporary store by those services.
+
+### Low-level Provider Options
 
 Flintrock exposes low-level provider options (e.g. [instance-initiated shutdown behavior](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/terminating-instances.html#Using_ChangingInstanceInitiatedShutdownBehavior)) so you can control the details of how your cluster is setup if you want.
 
--- a/flintrock.py
+++ b/flintrock.py
@@ -14,11 +14,6 @@ Major TODOs:
     * Check that EC2 enhanced networking is enabled.
 
 Other TODOs:
-    * Instance type <-> AMI type validation/lookup.
-        - Maybe this can be automated.
-        - Otherwise have a separate YAML file with this info.
-        - Maybe support HVM only. AWS seems to position it as the future.
-        - Show friendly error when user tries to launch PV instance type.
     * Use IAM roles to launch instead of AWS keys.
     * Setup and teardown VPC, routes, gateway, etc. from scratch.
     * Use SSHAgent instead of .pem files (?).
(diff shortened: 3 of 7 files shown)


nchammas:
@ericmjonas - I believe this PR patches up the last major functionality the AMPLab was looking for in Flintrock. Let me know what you think.

@shivaram, @broxtronix - What are your thoughts on the logic described above? For now, I don't have plans to add a separate `persistent-hdfs` module like spark-ec2, so I'm wondering if the way I'm targeting HDFS to work here covers your use cases well and meets your expectations.

nchammas:
I'll merge this in in the next few days if there's no feedback.

broxtronix:
Hey @nchammas.  I don't have too many comments on this one, because I tend to use ephemeral storage but not HDFS itself.  

However, with regards to the logic you describe above, I wonder if you could simplify things (and reduce the maintenance burden of supporting HDFS on many instance types) by simply saying that HDFS is only configured and available when there are ephemeral storage disks available.  That is, if a user starts up instances that don't have ephemeral storage, they don't get HDFS.

In practice, I suspect that the (EBS-backed) root drives are too small and too slow to really perform well as a backing store for HDFS anyway, and anyone seriously interested in running HDFS alongside spark will probably pick an instance type that has the locally-attached ephemeral storage they need to run HDFS efficiently.

That said, I guess one nice thing about the way that you have done it here is that people who are toying around with a small cluster can still expect Spark and HDFS to both be running.  Perhaps they have some very small tests they want to run that would require them both, and it would be annoying if they had to create a cluster with beefier instances just to try out HDFS.  So since you've implemented it this way, perhaps it is worth sticking with it for now.

nchammas:
Since the logic to switch between root and ephemeral storage is really just contained in [a couple of lines](https://github.com/nchammas/flintrock/pull/59/files#diff-1ec778e3c6cde832d96a7c08ba31f98eR139), I think it's fine to leave in. It doesn't add any complexity and, as you noted, it's helpful for people wanting to experiment with HDFS or do something small.

templates/hadoop/conf/hdfs-site.xml
```diff
@@ -9,6 +9,6 @@
 
   <property>
     <name>dfs.datanode.data.dir</name>
-    <value>/mnt/hdfs</value>
+    <value>{root_ephemeral_dirs}</value>
```
shivaram: I didn't follow the code very closely but will this just be `/media/ephemeral0' ? I think its a good idea to create `/media/ephemeral0/hdfs` and `/media/ephemeral0/tmp` and `/media/ephemeral0/spark` and use the appropriate variable in each place. The frameworks might take care of creating some directory structure but its always better to put them in different dirs in case they don't

nchammas: `{root_ephemeral_dirs}` will map to `/media/ephemeral0/hdfs,/media/ephemeral1/hdfs,...` or `/media/root/hdfs`, depending on whether the instance offers ephemeral storage.

Flintrock does that by [passing in the module](https://github.com/nchammas/flintrock/pull/59/files#diff-1ec778e3c6cde832d96a7c08ba31f98eR120) that the template is being formatted for, so each module gets its own space. So Spark's config files, for example, get `.../spark` for the same `{root_ephemeral_dirs}` template variable.

Does that answer your question?

shivaram: Yep - that sounds good to me.

nchammas:
Merging this in.

@ericmjonas - Let me know if you find any issues with this functionality.

PR #69, opened 2015-12-13
On failure, automatically retry Hadoop download a few times

### LINKED ISSUE #66: Downloading Hadoop from Apache mirrors is fragile
Ok, it was taking me forever to figure out why sometimes my cluster installs were hanging, appearing to never complete. It turns out that some of the different mirrors returned by the mirror checking script are incredibly slow (100kB/sec slow) . This has resulted in clusters taking up to 45min to launch. Is  there any way around this mirror search? It is extra complicated because different invocations of the mirror-finding script can return different mirrors on different ec2 nodes in a cluster, resulting in very erratic behavior.

nchammas:
Flintrock will now try up to 3 different Apache mirrors when downloading Hadoop, if necessary.

This won't address the issue of slow mirrors, unfortunately, but it does cover the common case of a bad mirror just failing to serve up the Hadoop file we're interested in.

Fixes #66.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@
 ### Added
 
 * Added `--assume-yes` option to the `launch` command. Use `--assume-yes` to tell Flintrock to automatically destroy the cluster if there are problems during launch.
+* Automatically retry Hadoop download from flaky Apache mirrors.
 
 
 ## 0.1.0 - 2015-12-11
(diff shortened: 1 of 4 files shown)


nchammas:
cc @ericmjonas @broxtronix

Let me know what you think of this. If it doesn't address the Hadoop download problems you are facing, then the other short-term fix we can look into is adding a download timeout.

This is difficult to do in Python, unfortunately, but we could probably shell out for the download itself and use GNU `timeout` in combination with `curl` to get the desired effect.

PR #77, opened 2016-01-25
Migrate from Paramiko + Threads to AsyncSSH

### LINKED ISSUE #9: Evaluate AsyncSSH as a replacement for Paramiko + asyncio thread executor
Right now Flintrock doesn't really use the full power of `asyncio`. We cheat by making regular SSH calls using Paramiko in `asyncio`'s [thread executor](https://github.com/nchammas/flintrock/blob/0e8a0624bcec08cefa3b199bd136be602b4720d8/flintrock.py#L579-L586). So all we get out of `asyncio` right now is perhaps a nicer API to the `threading` library.

The real reason Flintrock uses `asyncio` is that, as an orchestration tool, it manages remote resources. That means, fundamentally, Flintrock's main job is to do I/O against multiple servers across the network as quickly as possible. Most of Flintrock's interactions with the cluster nodes it manages are carried out over SSH. (The major exception is perhaps the work it does talking to the cloud provider.)

This use case sounds exactly like what `asyncio` is for.

[AsyncSSH](https://github.com/ronf/asyncssh#asyncssh-asynchronous-ssh-for-python) is a library that offers SSH over `asyncio`. It would replace both Paramiko as well as our use of the `asyncio` thread executor. More importantly, it promises to offer asynchronous SSH without the burden of threading.

I've had my eye on it for some time and have experimented with it on several occasions. The API is clean and the maintainer is active and responsive.

I want to seriously evaluate AsyncSSH as a replacement for Paramiko + thread executor, focusing on the following factors:
1. Does it make Flintrock easier to maintain, or at least not more complicated to maintain?
2. Does it make cluster operations, especially cluster launches, faster? How much faster?
3. What is the risk to depending on this relatively new library that does not yet have as much adoption or popularity as Paramiko?

nchammas:
This PR migrates Flintrock from Paramiko + Threads to AsyncSSH. The hope is that dropping the need for threads and fully utilizing asyncio will reduce Flintrock's memory footprint and, more importantly, speed up all cluster operations.
 
This is a work in progress. Remaining items:

- [x] Understand why launches are _slower_ with AsyncSSH, which is a surprising result.
- [x] Remove remaining, non-concurent uses of Paramiko, like for master configuration.
- [ ] Implement clean exit when the user hits Control + C during an operation.

Fixes #9.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 python:
-  - "3.4"
   - "3.5"
 install:
   - "pip install -r requirements/developer.pip"
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ That's not all. Flintrock has a few more [features](#features) that you may find
 
 Before using Flintrock, take a quick look at the [copyright](https://github.com/nchammas/flintrock/blob/master/COPYRIGHT) notice and [license](https://github.com/nchammas/flintrock/blob/master/LICENSE) and make sure you're OK with their terms.
 
-**Flintrock requires Python 3.4 or newer**, unless you are using one of our **standalone packages**. Flintrock has been thoroughly tested only on OS X, but it should run on all POSIX systems. We have plans to [add Windows support](https://github.com/nchammas/flintrock/issues/46) in the future, too.
+**Flintrock requires Python 3.5 or newer**, unless you are using one of our **standalone packages**. Flintrock has been thoroughly tested only on OS X, but it should run on all POSIX systems. We have plans to [add Windows support](https://github.com/nchammas/flintrock/issues/46) in the future, too.
 
 ### Release version
 
--- a/flintrock/core.py
+++ b/flintrock/core.py
@@ -1,5 +1,3 @@
-import asyncio
-import concurrent.futures
 import functools
 import json
 import os
@@ -9,8 +7,11 @@ import sys
 import time
 
 # Flintrock modules
-from .exceptions import SSHError, NodeError
-from .ssh import get_ssh_client, ssh_check_output, ssh
+from .ssh import (
+    get_ssh_client,
+    ssh_run,
+    ssh)
+from .util import run_against_hosts, sync_run
 
 FROZEN = getattr(sys, 'frozen', False)
 
@@ -134,21 +135,23 @@ class FlintrockCluster:
         """
         Start up all the services installed on the cluster.
 
-        This method assumes that the nodes constituting cluster were just
+        This method assumes that the nodes constituting the cluster were just
         started up by the provider (e.g. EC2, GCE, etc.) they're hosted on
         and are running.
         """
-        master_ssh_client = get_ssh_client(
-            user=user,
-            host=self.master_ip,
-            identity_file=identity_file)
+        master_ssh_client = sync_run(
+            get_ssh_client(
+                user=user,
+                host=self.master_ip,
+                identity_file=identity_file))
 
         with master_ssh_client:
-            manifest_raw = ssh_check_output(
-                client=master_ssh_client,
-                command="""
-                    cat /home/{u}/.flintrock-manifest.json
-                """.format(u=shlex.quote(user)))
+            manifest_raw = sync_run(
+                ssh_run(
+                    client=master_ssh_client,
+                    command="""
+                        cat /home/{u}/.flintrock-manifest.json
+                    """.format(u=shlex.quote(user))))
             # TODO: Reconsider where this belongs. In the manifest? We can implement
             #       ephemeral storage support as a Flintrock service, and add methods to
             #       serialize and deserialize critical service info like installed versions
@@ -158,14 +161,15 @@ class FlintrockCluster:
             #       instance type.
             # NOTE: As for why we aren't using ls here, see:
             #       http://mywiki.wooledge.org/ParsingLs
-            ephemeral_dirs_raw = ssh_check_output(
-                client=master_ssh_client,
-                command="""
-                    shopt -s nullglob
-                    for f in /media/ephemeral*; do
-                        echo "$f"
-                    done
-                """)
+            ephemeral_dirs_raw = sync_run(
+                ssh_run(
+                    client=master_ssh_client,
+                    command="""
+                        shopt -s nullglob
+                        for f in /media/ephemeral*; do
+                            echo "$f"
+                        done
+                    """))
 
         manifest = json.loads(manifest_raw)
         storage_dirs = StorageDirs(
@@ -180,7 +184,7 @@ class FlintrockCluster:
             service = globals()[service_name](**manifest)
             services.append(service)
 
-        partial_func = functools.partial(
+        start_node_partial = functools.partial(
             start_node,
             services=services,
             user=user,
@@ -188,12 +192,13 @@ class FlintrockCluster:
             cluster=self)
         hosts = [self.master_ip] + self.slave_ips
 
-        _run_asynchronously(partial_func=partial_func, hosts=hosts)
+        run_against_hosts(async_partial_func=start_node_partial, hosts=hosts)
 
-        master_ssh_client = get_ssh_client(
-            user=user,
-            host=self.master_ip,
-            identity_file=identity_file)
+        master_ssh_client = sync_run(
+            get_ssh_client(
+                user=user,
+                host=self.master_ip,
+                identity_file=identity_file))
 
         with master_ssh_client:
             for service in services:
@@ -254,14 +259,14 @@ class FlintrockCluster:
         else:
             target_hosts = [self.master_ip] + self.slave_ips
 
-        partial_func = functools.partial(
+        run_command_node_partial = functools.partial(
             run_command_node,
             user=user,
             identity_file=identity_file,
             command=command)
         hosts = target_hosts
 
-        _run_asynchronously(partial_func=partial_func, hosts=hosts)
+        run_against_hosts(async_partial_func=run_command_node_partial, hosts=hosts)
 
     def copy_file_check(self):
         """
@@ -291,7 +296,7 @@ class FlintrockCluster:
         else:
             target_hosts = [self.master_ip] + self.slave_ips
 
-        partial_func = functools.partial(
+        copy_file_node_partial = functools.partial(
             copy_file_node,
             user=user,
             identity_file=identity_file,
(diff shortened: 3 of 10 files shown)


nchammas:
I expected Flintrock to be generally faster with AsyncSSH as compared to Paramiko + Threads, especially at launching large clusters.

What I'm seeing goes against my expectations. When launching a cluster with 100 slaves, Flintrock is actually _slower_ with AsyncSSH:
* Paramiko + Threads: 5 minutes 15 seconds
* AsyncSSH: 5 minutes 40 seconds

These results roughly hold across repeated tries with clusters of different sizes.

My expectations may just be wrong, but I suspect there is a problem with Flintrock that makes things go slower than they should.

I ran `cProfile` against both versions of Flintrock, but I'm not sure how to interpret the results.

Paramiko + Threads:
```
Sun Jan 24 17:43:31 2016    flintrock-threads.prf

         2519940 function calls (2366591 primitive calls) in 318.663 seconds

   Ordered by: internal time
   List reduced from 4172 to 10 due to restriction <10>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      205  178.395    0.870  178.395    0.870 {method 'control' of 'select.kqueue' objects}
       11   67.036    6.094   67.036    6.094 {built-in method time.sleep}
     2910   50.858    0.017   50.858    0.017 {method 'read' of '_ssl._SSLSocket' objects}
     2232   11.292    0.005   11.292    0.005 {method 'acquire' of '_thread.lock' objects}
        8    2.992    0.374    2.992    0.374 {method 'connect' of '_socket.socket' objects}
        5    2.171    0.434    2.171    0.434 {method 'do_handshake' of '_ssl._SSLSocket' objects}
        5    1.636    0.327    1.639    0.328 {built-in method _socket.gethostbyname}
       14    0.466    0.033    0.466    0.033 {method 'recv_into' of '_socket.socket' objects}
       23    0.347    0.015    0.347    0.015 {method 'feed' of 'xml.etree.ElementTree.XMLParser' objects}
 18250/16    0.188    0.000    1.572    0.098 .../venv/lib/python3.5/site-packages/botocore/parsers.py:281(_handle_structure)
```

AsyncSSH:
```
Sun Jan 24 17:25:53 2016    flintrock-asyncio.prf

         16214051 function calls (16035212 primitive calls) in 343.101 seconds

   Ordered by: internal time
   List reduced from 5299 to 10 due to restriction <10>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    11346  164.515    0.014  164.515    0.014 {method 'control' of 'select.kqueue' objects}
       10   64.022    6.402   64.022    6.402 {built-in method time.sleep}
     2645   41.517    0.016   41.517    0.016 {method 'read' of '_ssl._SSLSocket' objects}
     6753   34.381    0.005   34.381    0.005 {method 'acquire' of '_thread.lock' objects}
    25449    4.409    0.000   14.131    0.001 .../lib/python3.5/traceback.py:313(extract)
        8    4.072    0.509    4.072    0.509 {built-in method _socket.getaddrinfo}
        5    3.422    0.684    3.422    0.684 {method 'do_handshake' of '_ssl._SSLSocket' objects}
   395843    2.901    0.000    2.901    0.000 {built-in method posix.stat}
      649    2.022    0.003    2.022    0.003 {method 'connect' of '_socket.socket' objects}
   828407    1.148    0.000    3.484    0.000 .../lib/python3.5/traceback.py:279(line)
```

This is my first time profiling an asyncio application (or, really, any Python application), so I'm wondering how I should approach this. My goal is to understand why, exactly, Flintrock is slower with AsyncSSH and, if possible, fix any mistakes I've made in my use of it.

nchammas:
Hey @ronf, are there any obvious things I should look into that, from your experience, may result in the slowdown I'm seeing here?

ronf:
Is that reporting wall-clock time or total CPU time used? One thing to realize about asyncio is that it's only going to use one core on a multi-core system, whereas thread-based code can use multiple cores (though not always all that efficiently given the Python global interpreter lock). This could result in code taking long in wall-clock time if you only have a single event loop running vs. a threaded version.

If your workload is parallelizable, you could of course start multiple event loops and run on them different cores in parallel, or use multiple processes to achieve the same result if you have no state to share and want to avoid any Python global interpreter lock issues, but that may not be consistent with what you're trying to measure here.

nchammas:
The 5:15 vs 5:40 is wall clock time.

Flintrock is essentially an orchestration tool. It spins up a bunch of instances on Amazon's EC2 (that's always a single API call, regardless of how many instances are requested) and then sends commands and files to those instances via SSH to install and setup stuff on each instance in parallel.

So Flintrock doesn't really do any work locally. All the real work is done by those remote instances, and Flintrock's job is to manage them via SSH.

From what I understand, this seems like the ideal use case for asyncio: Doing network I/O against many remote resources concurrently. I don't think I need multiple event loops or multiple cores because, again, the Flintrock client isn't doing much more than sending and receiving data across the network.

So my basic assumption here is: If Flintrock is launching, say, a 100-node cluster, then asyncio's event loop will be faster than 100 threads competing with each other to send data over SSH.

At a high level, is this assumption valid? I'm wondering if it's worth investigating the slowdown documented here or if, perhaps, I just started out with the wrong set of assumptions.

broxtronix:
Hey @nchammas.  My experience has been that SSH can be moderately CPU intensive to set up a new SSH connection.  In fact, I have sometimes needed to introduce a brief, random delay before opening SSH connection so that I don't slam all local cores at the same time and grind my local machine to a halt as it tries to open 100 outgoing SSH connections.  And although communicating over an open SSH connection does take negligible resources, these resources can add up quickly when managing 100+ open connections at the same time.    

Based on this, I suspect the AsyncSSH event loop is a little bit overwhelmed at times.  Even if connection opening is carefully made to be asynchronous, with mulitple re-entry points into the event loop during the connection negotiation process, there is still only a single CPU left to negotiate all of these connections, and that is going to take longer than a multithreaded approach.  Using threads and allowing multiple CPU cores to participate in the opening and maintenance of SSH connections is likely a little faster than going with an event loop running on a single core.  But in that case you just have to be sure not to use *all* of the local machine's resources at once!

nchammas:
> My experience has been that SSH can be moderately CPU intensive to set up a new SSH connection.

Interesting point. I'm guessing the cryptography methods that are involved in setting up and maintaining an SSH connection are compute-intensive.

So this raises the question: How can I validate this hypothesis that CPU contention -- as opposed to something else, like unnecessary blocking -- is what is slowing launches down?

My profiling skills are still quite elementary, so I'm hoping to pick up some performance debugging techniques as part of this analysis. If you have any tips to share, please do!

broxtronix:
I'm not sure if I have an silver bullet tips, but I can try to point you in a good general direction.  It looks like the asyncio event loop has a [debug mode](https://docs.python.org/3/library/asyncio-dev.html).  I wonder if you can turn something like this on for AsyncSSH.  You can infer a lot if you can see when and how the event loop is executing short bits of code.  Big delays may show up in those logs.

On OSX, the "Instruments.app" (which is part of XCode) is a really excellent time profiler, and can be used to monitor the entire system.  It puts everything in context, and can be helpful for getting more insight.

On linux and OSX, `htop` is honestly a good place start in building an intuition about what is going on.  Even a short burst of CPU activity should show up as a momentary blip in `htop` and it should also be clear when one CPU is tied up for a long time.  It's a coarse monitoring tool, but like good ol fashioned `printf()` debugging, I usually can't get by without it. 

Finally, [this page](http://www.brendangregg.com/linuxperf.html) is a pretty fun read if you want to bone up on your general Linux debugging and system diagnostics skills.  The list of debugging tools there is exhaustive!

ronf:
If the benchmark you have is a command line tool, you can run the "time" command to determine not only the wall-clock time but also the total CPU time used. That will give you an idea whether or not the too is actually CPU intensive enough that multiple cores may be helping the Paramiko solution. Just look for the total CPU time used being close to or greater than the wall clock time.

So far, I haven't build any performance benchmarking into AsyncSSH. I did do some tests early on of some large file transfers vs. the C version of OpenSSH and was pleased that it ran at about 70% of the speed (better than I expected given that all packet processing was pure Python code and only the low-level crypto was in C). That was just opening one connection and spending most of the time doing symmetric encryption, which isn't as expensive as opening a lot of short lived connections. Also, it is still single-threaded and I knew that would be a bottleneck if you ran enough connections in a single event loop. I'm just not sure how many you'd need before that was an issue (and of course that would depend on the machine you're on and the crypto algorithms you choose).

In general, I think you're right that AsyncSSH has the potential to be faster by running all in a single thread in cases where it is spending most of its time blocked on I/O and not CPU-bound.

I did a quick test here just now. On my iMac (4 GHz Skylake i7 processor) opening 100 connections to localhost in parallel running a single simple command, I only managed to use about 30-40% of a single CPU. One example run showed:

./multi_client.py  4.22s user 0.21s system 38% cpu 11.394 total

(so 11.394 seconds of wall clock time, with 4.43 seconds of total CPU time, or 38%)

Opening a larger number of connections in parallel started to hit other limits on the system like the number of open file descriptors available both to the AsyncSSH process and other things it was communicating with like ssh-agent.

nchammas:
I haven't been able to dive too deeply into this issue -- been busy getting Flintrock 0.2 out! -- but I did manage to run a quick test with `time` as @ronf suggested.

I launched several clusters with 50 slaves. Here are best times for each version of Flintrock:

Paramiko + Threads:
```
real  2m48.763s
user  0m22.134s
sys   0m2.829s
```

AsyncSSH:
```
real  2m30.410s
user  0m9.189s
sys   0m1.774s
```

This time the AsyncSSH version of Flintrock was faster, which contradicts my results from the start of the thread. I think this is more a testament to the variability of AWS than to anything about asyncio vs. threads.

The more interesting result is the user CPU time (`user`), which confirms the obvious -- that Flintrock is I/O bound -- but which also shows that AsyncSSH is less CPU-intensive than Paramiko.

This isn't that enlightening, but I think it suggests that CPU is not the issue here. In fact, it may be that my [original test results](https://github.com/nchammas/flintrock/pull/77#issuecomment-174712063) were totally invalid (even though I ran that test several times, painful as it was for my wallet) and AWS variability trumps any performance difference between Paramiko + Threads and AsyncSSH.

Anyway, further research is required.™

nchammas:
Just FYI: This PR is blocked pending a resolution to https://github.com/pyca/cryptography/issues/2749, which prevents Flintrock from exiting cleanly when the user interrupts it with `Control + C` at key points.

nchammas:
I'll have to close this PR for now and revisit it sometime in the future. I'm still very interested in migrating Flintrock to AsyncSSH and Python 3.5's `async`/`await` syntax, but last I tried I wasn't able to overcome the segfaults I was getting from Cryptography / cffi.

Perhaps after 0.5 or 0.6 is out I'll give this another shot. Hopefully by then the segfaults will be gone and I'll get that fix for "free".

PR #98, opened 2016-03-20
Allow users to set the git commit to install Spark from as 'latest'

### LINKED ISSUE #61: Allow users to set the git commit to install Spark from as 'latest'
Oftentimes, I imagine users wanting to install Spark from git may just care about getting the latest version on master to test a bug fix or new feature.

In that case, it's a hassle to have to look up the latest commit when Flintrock can just do it for you. Set the desired commit to `latest` and Flintrock will lookup the latest commit automatically.

There are 3 approaches we can take to implement a convenience feature like this:

1. Import some Python library for interacting with git and use that to query the latest version. Advantage: No implicit user dependencies. Disadvantage: It's another Flintrock dependency we have to vet.
2. Use `git` at the command line as such:

  ```
git ls-remote https://github.com/apache/spark master
  ```
  Advantage: Very simple. Disadvantage: We've created an implicit dependency on `git` being installed on the machine running Flintrock.
3. Query the GitHub API. Advantage: Still simple. Disadvantage: Now we have a dependency on GitHub for something that is strictly git functionality. This may not be a big deal since Spark itself, and likely most forks of it, are hosted on GitHub.

Depends on #25.

BenFradet:
fixes #61

### DIFF
--- a/flintrock/flintrock.py
+++ b/flintrock/flintrock.py
@@ -320,11 +320,12 @@ def get_last_commit_sha(git_repository: str):
             result = json.loads(response.readall().decode('utf-8'))
             return result[0]['sha']
         else:
-            raise UsageError('Error: Error retrieving last commit\'s SHA of {repo}.\n'
-                             '  {code} {message}'.format(
-                                repo=git_repository,
-                                code=response.getcode(),
-                                message=response.msg))
+            raise UsageError(
+                'Error: Error retrieving last commit\'s SHA of {repo}.\n'
+                '  {code} {message}'.format(
+                    repo=git_repository,
+                    code=response.getcode(),
+                    message=response.msg))
 
 
 @cli.command()
--- a/flintrock/flintrock.py
+++ b/flintrock/flintrock.py
@@ -176,7 +176,8 @@ def cli(cli_context, config, provider):
               help="Spark release version to install.")
 @click.option('--spark-git-commit',
               help="Git commit hash to build Spark from. "
-                   "Can be 'latest' to build Spark from the latest available commit. "
+                   "Can be 'latest' to build Spark from the latest available commit on the "
+                   "default branch. "
                    "--spark-version and --spark-git-commit are mutually exclusive.",
               default="latest",
               show_default=True)
@@ -308,7 +309,7 @@ def launch(
 
 def get_latest_commit(github_repository: str):
     """
-    Retrieve the last commit's SHA from a repository hosted on Github.
+    Retrieve the latest commit on the default branch of a repository hosted on Github.
     """
     parsed_url = urllib.parse.urlparse(github_repository)
     if parsed_url.netloc != 'github.com':
--- a/flintrock/flintrock.py
+++ b/flintrock/flintrock.py
@@ -316,16 +316,8 @@ def get_last_commit_sha(git_repository: str):
     owner_repo = parsed_url.path[:-4] if parsed_url.path.endswith(".git") else parsed_url.path
     url = "https://api.github.com/repos{owner_repo}/commits".format(owner_repo=owner_repo)
     with urllib.request.urlopen(url) as response:
-        if response.getcode() == 200:
-            result = json.loads(response.readall().decode('utf-8'))
-            return result[0]['sha']
-        else:
-            raise UsageError(
-                'Error: Error retrieving last commit\'s SHA of {repo}.\n'
-                '  {code} {message}'.format(
-                    repo=git_repository,
-                    code=response.getcode(),
-                    message=response.msg))
+        result = json.loads(response.read().decode('utf-8'))
+        return result[0]['sha']
 
 
 @cli.command()
--- a/flintrock/flintrock.py
+++ b/flintrock/flintrock.py
@@ -1,10 +1,13 @@
 import os
 import posixpath
 import errno
+import json
 import resource
 import sys
 import shutil
 import textwrap
+import urllib.parse
+import urllib.request
 import warnings
 
 # External modules
@@ -172,8 +175,11 @@ def cli(cli_context, config, provider):
 @click.option('--spark-version',
               help="Spark release version to install.")
 @click.option('--spark-git-commit',
-              help="Git commit hash to build Spark from. "
-                   "--spark-version and --spark-git-commit are mutually exclusive.")
+              help="Git commit hash to build Spark from."
+                   "Can be \"latest\" to build Spark from the latest available commit."
+                   "--spark-version and --spark-git-commit are mutually exclusive.",
+              default="latest",
+              show_default=True)
 @click.option('--spark-git-repository',
               help="Git repository to clone Spark from.",
               default='https://github.com/apache/spark.git',
@@ -295,6 +301,22 @@ def launch(
             instance_initiated_shutdown_behavior=ec2_instance_initiated_shutdown_behavior)
     else:
         raise UnsupportedProviderError(provider)
+
+
+def get_last_commit_sha(git_repository: str):
+    """
+    Retrieve the last commit's SHA from a repository hosted on Github.
+    """
+    parsed_url = urllib.parse.urlparse(git_repository)
+    if parsed_url.netloc != "github.com":
+        raise UsageError('Error: Retrieving the last commit\'s SHA only works with '
+                         'repositories hosted on Github.')
+    owner_repo = parsed_url.path[:-4] if parsed_url.path.endswith(".git") else parsed_url.path
+    url = "https://api.github.com/repos{owner_repo}/commits".format(owner_repo=owner_repo)
+    with urllib.request.urlopen(url) as response:
+        if response.getcode() == 200:
+            result = json.loads(response.readall().decode('utf-8'))
+            return result[0]['sha']
 
 
 @cli.command()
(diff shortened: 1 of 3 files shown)


flintrock/flintrock.py
```diff
@@ -297,6 +305,21 @@ def launch(
...
+    """
+    parsed_url = urllib.parse.urlparse(git_repository)
+    if parsed_url.netloc != "github.com":
+        raise UsageError('Error: Retrieving the last commit\'s SHA only works with '
+                         'repositories hosted on Github.')
+    owner_repo = parsed_url.path[:-4] if parsed_url.path.endswith(".git") else parsed_url.path
+    url = "https://api.github.com/repos{owner_repo}/commits".format(owner_repo=owner_repo)
+    with urllib.request.urlopen(url) as response:
+        result = json.loads(response.read().decode('utf-8'))
+        return result[0]['sha']
```
BenFradet: @nchammas I'd like to get your input on how to best handle the failures.

flintrock/flintrock.py
```diff
@@ -172,8 +175,11 @@ def cli(cli_context, config, provider):
 @click.option('--spark-version',
               help="Spark release version to install.")
 @click.option('--spark-git-commit',
-              help="Git commit hash to build Spark from. "
-                   "--spark-version and --spark-git-commit are mutually exclusive.")
+              help="Git commit hash to build Spark from."
+                   "Can be \"latest\" to build Spark from the latest available commit."
```
nchammas: The trailing space I had in there previously was intentional. ;)

These strings need trailing spaces, otherwise when they are joined each sentence will be glued to the previous one.

flintrock/flintrock.py
```diff
@@ -172,8 +175,11 @@ def cli(cli_context, config, provider):
...
+                   "--spark-version and --spark-git-commit are mutually exclusive.",
+              default="latest",
```
nchammas: Nitpick: Flintrock follows the pattern of single quotes for programmatically meaningful values, and double quotes for English prose. So throughout this PR, I want us to use `'latest'` instead of `"latest"`.

The exception to this is when naming command-line options, to follow the convention set there by Click.

flintrock/flintrock.py
```diff
@@ -297,6 +305,21 @@ def launch(
         raise UnsupportedProviderError(provider)
 
 
+def get_last_commit_sha(git_repository: str):
```
nchammas: I think this signature is better expressed as follows:

```python
def get_latest_commit(github_repository: str):
```

The `_latest_` matches the name of the option value `'latest'`, and the `github_repository` emphasizes that this method is not for any git repository.

flintrock/flintrock.py
```diff
@@ -268,6 +274,8 @@ def launch(
         elif spark_git_commit:
             print("Warning: Building Spark takes a long time. "
                   "e.g. 15-20 minutes on an m3.xlarge instance on EC2.")
+            if spark_git_commit == "latest":
+                spark_git_commit = get_last_commit_sha(spark_git_repository)
```
nchammas: Here is a good place to tell the user what commit we're building Spark at, otherwise they will not know without logging into the cluster and running the appropriate git command.

flintrock/flintrock.py
```diff
@@ -297,6 +305,21 @@ def launch(
         raise UnsupportedProviderError(provider)
 
 
+def get_last_commit_sha(git_repository: str):
+    """
+    Retrieve the last commit's SHA from a repository hosted on Github.
+    """
+    parsed_url = urllib.parse.urlparse(git_repository)
+    if parsed_url.netloc != "github.com":
+        raise UsageError('Error: Retrieving the last commit\'s SHA only works with '
```
nchammas: Nit: As noted before, I prefer this string be encapsulated by double quotes since it is prose, and the opposite goes for `"github.com"`.

Also, I'm not totally sure about the terminology, but I think "the latest commit" is sufficient. No need to mention SHA.

flintrock/flintrock.py
```diff
@@ -297,6 +305,21 @@ def launch(
...
+                         'repositories hosted on Github.')
+    owner_repo = parsed_url.path[:-4] if parsed_url.path.endswith(".git") else parsed_url.path
```
nchammas: This is a nice touch.

Actually, it makes me wonder: Does everything work if we drop the `.git` from the default repo name? I noticed the config template includes it, and I forget if that was a necessity. If not, it's better to remove it.

flintrock/flintrock.py
nchammas: Hmm, I think this is mostly fine as-is since with "unexpected" failures like this we simply want the error to bubble up to the user so they can debug.

The only problem I see here is that a misleading index or lookup error (from the `[0]['sha']`) may conceal the underlying problem with getting the commit (e.g. the request "succeeds" but returns bad info). Not sure if there is a simple way to separate those two problems.

tests/test_flintrock.py
```diff
@@ -9,7 +9,12 @@
     option_name_to_variable_name,
     variable_name_to_option_name,
     option_requires,
-    mutually_exclusive
+    mutually_exclusive,
+    get_last_commit_sha
+)
+
+from urllib.error import (
```
nchammas: Nit: I generally reserve `from ... import` for Flintrock objects. Exceptions include datatypes like `namedtuple` and `datetime`.

tests/test_flintrock.py
```diff
@@ -126,3 +131,17 @@ def test_mutually_exclusive():
                 '--option1',
                 '--option2'],
             scope=locals())
+
+
+def test_get_last_commit_sha():
+    sha1 = get_last_commit_sha("https://github.com/apache/spark.git")
+    assert len(sha1) == 40
+
+    sha2 = get_last_commit_sha("https://github.com/apache/spark.git")
```
nchammas: What's the purpose of this second test?

nchammas:
Thanks for working on this @BenFradet! It looks good, and I appreciate that you added tests for the new method too. I went through a left a bunch of comments.

One last thing we should address here is what branch gets queried when we ask for the latest commit. I believe GitHub (or maybe Git?) has a concept of a default branch for a repo, so your call to the GitHub API probably translates to "the latest commit on the default branch". For some repos it may not be `master`.

We should just clarify this with a few extra words in the help string.

BenFradet:
@nchammas thanks for your review.
Will make the necessary changes later today.

flintrock/flintrock.py
BenFradet: Would it be a good idea to include the short version of the commit message as well?

flintrock/flintrock.py
BenFradet: Both (with and without `.git`) work as a remote.

flintrock/flintrock.py
nchammas: That would be a nice touch, but strictly optional at this point in my view. Would you find it useful?

flintrock/flintrock.py
nchammas: Great. Perhaps then we should update the config template and defaults to remove the `.git`. We could probably even remove the logic that strips the `.git`, just to simplify things. Every little bit counts!

tests/test_flintrock.py
BenFradet: I wanted to remote the `.git` and forgot.

flintrock/flintrock.py
BenFradet: I would find it more useful than just the sha if I needed to go trough the history on git/github to find which commit had been checked out.
But since the point of this PR is not to care about the commit being checked out, I don't really now.

nchammas: I'd say leave it out for starters and keep things simple. We can easily add it in later. :+1: 

And I agree: If you care about the commit, you probably already looked it up beforehand.

flintrock/flintrock.py
BenFradet: Do you mean in this PR.

nchammas: Yup, as part of this PR simply remove all mention of `.git` in the repo name if we don't need it, and remove the logic to strip it.

flintrock/flintrock.py
BenFradet: Do you mean that the `UsageError`s should be double quoted string as well?
Because throughout the file they're single quoted.

nchammas: Oh, you're right. I did that because I was double-quoting option names to be consistent with Click. 

Hmm, never mind then. I've been inconsistent in this case; I'll fix it myself later.

BenFradet:
@nchammas I think I've addressed all your comments!

nchammas:
Thanks @BenFradet! This looks great. I'm going to pull it in to a new PR and make some tweaks before merging it in. Thank you for working on this and going through the rounds of review. :+1:

BenFradet:
np

flintrock/flintrock.py
```diff
@@ -173,10 +176,14 @@ def cli(cli_context, config, provider):
               help="Spark release version to install.")
 @click.option('--spark-git-commit',
               help="Git commit hash to build Spark from. "
-                   "--spark-version and --spark-git-commit are mutually exclusive.")
+                   "Can be 'latest' to build Spark from the latest available commit on the "
+                   "default branch. "
+                   "--spark-version and --spark-git-commit are mutually exclusive.",
+              default="latest",
```
nchammas: FYI: I am going to remove this default setting because it breaks users who have configured `--spark-version` in their config file. I may have instructed you to set this default; apologies.

BenFradet: sure, no problem

PR #99, opened 2016-03-22
Allow users to set the git commit to install Spark from as 'latest'

### LINKED ISSUE #61: Allow users to set the git commit to install Spark from as 'latest'

nchammas:
Made some minor tweaks to @BenFradet's work.

Supersedes / Fixes #98.
Fixes #61.

### DIFF
--- a/tests/test_flintrock.py
+++ b/tests/test_flintrock.py
@@ -137,5 +137,5 @@ def test_get_latest_commit():
     with pytest.raises(UsageError):
         get_latest_commit("https://google.com")
 
-    with pytest.raises(urllib.error.HTTPError):
-        get_latest_commit("https://github.com/apache/spark2")
+    with pytest.raises(Exception):
+        get_latest_commit("https://github.com/apache/nonexistent-repo")
--- a/tests/test_flintrock.py
+++ b/tests/test_flintrock.py
@@ -1,5 +1,6 @@
 # External modules
 import pytest
+import urllib.error
 
 # Flintrock modules
 from flintrock.exceptions import (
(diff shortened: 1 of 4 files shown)


tests/test_flintrock.py
```diff
@@ -1,5 +1,6 @@
 # External modules
 import pytest
+import urllib.error
```
BenFradet: @nchammas This import isn't used anymore

nchammas: Good catch. I'll remove it.

nchammas: Done: 73bf8ba7396ff5c35a5f6a13d57c7a7432bd4b6c

Issue #71, opened 2015-12-15
Add option to download Hadoop from a custom URL


nchammas:
As a follow-up to the discussion in #66, perhaps we should add an option to let users download Hadoop from a custom URL.

* Command-line option: `--hdfs-download-source`
* Config file option:

  ```yaml
modules:
  hdfs:
    version: 2.7.1
    download-source: "http://www.apache.org/dyn/closer.lua/hadoop/common/hadoop-{v}/hadoop-{v}.tar.gz?as_json"
```

`{v}` will be replaced by the HDFS version passed in to Flintrock, either via the config file or via the command line.

The `.../dyn/closer.lua/...` value will be Flintrock's internal default, which the user can replace with a specific Apache mirror, or some other source. The only requirements are that the package be downloadable from the cluster without authentication, and that the URL contain the `{v}` template somewhere.

I would like to deprecate this option as soon as we have a more robust way of downloading Hadoop quickly and reliably, since that's the main motivation for adding this option in.

Related to #66, #69, #84.

nchammas:
@ericmjonas - To continue our discussion from #84 about an Apache mirror blacklist here, what mirrors have you noticed take forever to serve up Hadoop?

The mirrors I currently have on my list are:
* 104.45.233.178 
* 74.206.97.82

ericmjonas:
I have to be honest, I haven't actually cataloged them yet -- I normally
get so frustrated that I abort and immediately go to hacking
download_hadoopy.py. I'm sorry :(

nchammas:
OK, so we can say that at least for you, the download source option described here would work really well. Right? :grinning: 

Perhaps we should get that working first, and then later revisit the idea of a mirror blacklist, or some kind of speculative download-retry mechanism for when a download seems like it it's gonna take forever.

What do you say?

ericmjonas:
On Sat, Feb 20, 2016 at 9:59 AM, Nicholas Chammas <notifications@github.com>
wrote:

> What do you say?
>

That sounds great!

nchammas:
I agree with you that the out-of-box experience is critical. For the record, I'm hesitant to dive into these other approaches because:
* Speculative download-retry: This would be the "ultimate" solution, but it seems like it would cost a lot in terms of complexity. We'd need a way to monitor the download rate, kill it based on some heuristic, and then try a new download. Seems like a lot of work to me for a relatively limited feature.
[...]

PR #100, opened 2016-03-27
Make download-hadoop.py Python 2/3 compatible

nchammas:
This should be the last change required to get Flintrock to work with AMIs that have Python 3 as the default Python.

I haven't tested this with a full launch. I've just tested the download script individually from my laptop to make sure it works with both Python 2 and 3.


nchammas:
FYI @BenFradet, since you may be working on the script as part of #71.

BenFradet:
noted, thanks

Issue #102, opened 2016-04-06
Allow the user to connect to the master on port 7077


rmessner:
To be able to connect a zeppelin or spark-shell to my remote cluster, I had to manually update the security group to allow the port 7077 from my IP.

Would be nice that flintrock set it up by default.

nchammas:
Fixed by #103.

PR #103, opened 2016-04-06
add security rule : open port 7077 from client 

rmessner:
Open the port 7077, so clients ( e.g. zeppelin ) can connect to the spark master like described in the issue #102

### DIFF
--- a/flintrock/ec2.py
+++ b/flintrock/ec2.py
@@ -367,6 +367,12 @@ def get_or_create_ec2_security_groups(
             from_port=4040,
             to_port=4040,
             cidr_ip=flintrock_client_cidr,
+            src_group=None),
+        SecurityGroupRule(
+            ip_protocol='tcp',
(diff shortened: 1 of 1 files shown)


nchammas:
Thanks for working on this.

Is there a way I can test this PR is working as expected without installing Zeppelin? Both `curl` and `telnet` against port 7077 are refused, even if I try from the cluster.

nchammas:
Alternately, if you can tell me that you've tested this both before and after this PR, and that this fixes the issue you had with connecting local Zeppelin to the remote Spark master, then that's good enough for me too. :smile:

rmessner:
I used spark-shell and zeppelin to test it and it worked well.

The only limitation is that you can't be behind a NAT ( except if you forward the 4040 port to your laptop )

nchammas:
OK great, thanks!

Issue #88, opened 2016-03-01
Cannot install package pre-built for Hadoop 2.4


hhbyyh:
There's no way to config which pre-built version to install for Spark 1.6.0.

nchammas:
Correct, though you could easily hack this by modifying [this line](https://github.com/nchammas/flintrock/blob/1e582dcf7a2fb3912343e93ee4dac6c297f920cf/flintrock/services.py#L220-L221).

One of Flintrock's [anti-use cases](https://github.com/nchammas/flintrock#launching-out-of-date-services) is support for out-of-date versions of services like Hadoop. I've considered allowing users to specify the distribution of Spark to install, but I haven't found a good motivating use case for it. Generally, Flintrock targets the latest versions of everything.

What's your use case for wanting Spark built against Hadoop 2.4?

hhbyyh:
We want to save the compute result to S3. Yet Hadoop 2.6 has problem with that.
You may check https://issues.apache.org/jira/browse/SPARK-7442 for more details.
Generally speaking, pre-built for Hadoop 2.4 can perfectly support writing to S3n directly while 2.6 cannot.

nchammas:
Ah yes, that issue. (Funny, I was actually the one who reported it!)

I have a TODO on my list to deploy the additional JARs required to get Spark / Hadoop 2.6 working with S3, though I didn't create an issue for it here on GitHub yet.

If Flintrock automatically did that for you, would you still be interested in deploying Spark pre-built for Hadoop 2.4?

Alternately, I can expose the Spark profile as an option to get around this problem until [SPARK-7481](https://issues.apache.org/jira/browse/SPARK-7481) is resolved.

ereed-tesla:
Alternatively, you can get the packages at runtime with (e.g.):
`pyspark --packages com.amazonaws:aws-java-sdk-pom:1.10.34,org.apache.hadoop:hadoop-aws:2.6.0`

This allows you to access S3n.

nchammas:
Thanks for the tip @ereed-tesla! That'll be handy until we find a good solution for this issue, whether here or upstream.

hhbyyh:
@ereed-tesla Thanks for the response. Do you happen to know how to deal with it if I'm using spark-submit.

hhbyyh:
@nchammas All I want is to write to S3....

ereed-tesla:
The same arg (--packages) works for spark-submit.

nchammas:
[...]

PR #104, opened 2016-04-06
Add a new option for an alternate mirror for spark binaries

### LINKED ISSUE #101: Add option to download Spark from a custom URL
We are trying to setup a cluster with Spark 1.6.1, without HDFS, but it fails.
The reason is that the file downloaded is corrupted, so we can't untar it.

Would be a nice feature to allow user to specify an alternate mirror to download the pre-build spark. This would add a new key to configuration, like : 

```
services:
  spark:
    version: 1.6.0
    preferred-mirror: http://apache.crihan.fr/dist/spark/spark-${spark_version}/${file}
```

The default value would be `https://s3.amazonaws.com/spark-related-packages/${file}` and the variables available `file` ( as the file name of pre-build spark is always the same whatever the mirror is ), `spark_version` ( e.g. : 1.6.1) , `distribution` ( e.g. : hadoop2.6 )

Related to #71.

rmessner:
Fixes #101.

### DIFF
--- a/flintrock/config.yaml.template
+++ b/flintrock/config.yaml.template
@@ -1,6 +1,7 @@
 services:
   spark:
     version: 1.6.0
+    # preferred-mirror: # optional; default to 'https://s3.amazonaws.com/spark-related-packages/${file}'
     # git-commit: latest  # if not 'latest', provide a full commit SHA; e.g. d6dc12ef0146ae409834c78737c116050961f350
     # git-repository:  # optional; defaults to https://github.com/apache/spark
   hdfs:
--- a/flintrock/config.yaml.template
+++ b/flintrock/config.yaml.template
@@ -1,7 +1,8 @@
 services:
   spark:
     version: 1.6.0
-    # preferred-mirror: # optional; default to 'https://s3.amazonaws.com/spark-related-packages/${file}'
+    # distribution: # optional; default to '2.6'
+    # download-source: # optional; default to 'https://s3.amazonaws.com/spark-related-packages/spark-${version}-bin-hadoop${distribution}.tgz'
     # git-commit: latest  # if not 'latest', provide a full commit SHA; e.g. d6dc12ef0146ae409834c78737c116050961f350
     # git-repository:  # optional; defaults to https://github.com/apache/spark
   hdfs:
--- a/flintrock/flintrock.py
+++ b/flintrock/flintrock.py
@@ -186,6 +186,11 @@ def cli(cli_context, config, provider):
               help="Git commit to build Spark from. "
                    "Set to 'latest' to build Spark from the latest commit on the "
                    "repository's default branch.")
+@click.option('--spark-preferred-mirror',
+              help="HTTP mirror to download the spark binaries. "
+                   "Available variable : file, spark_version, distribution",
+              default="https://s3.amazonaws.com/spark-related-packages/${file}",
+              show_default=True)
 @click.option('--spark-git-repository',
               help="Git repository to clone Spark from.",
               default='https://github.com/apache/spark',
@@ -222,6 +227,7 @@ def launch(
         spark_version,
         spark_git_commit,
         spark_git_repository,
+        spark_preferred_mirror,
         assume_yes,
         ec2_key_name,
         ec2_identity_file,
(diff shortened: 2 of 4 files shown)


nchammas:
@rmessner thank you for tackling this. As you experienced, it is useful in times when, for whatever reason, the default Spark packages on S3 that Flintrock uses are corrupt.

We will probably want to use this pattern here to also solve #71. As such, I'd like the option name and URL template to be consistent with what's proposed there. Is that OK with you?

flintrock/config.yaml.template
```diff
@@ -1,6 +1,7 @@
 services:
   spark:
     version: 1.6.0
+    # preferred-mirror: # optional; default to 'https://s3.amazonaws.com/spark-related-packages/${file}'
```
nchammas: As proposed in #71, I prefer `{v}` or even `{version}` instead of `${file}`, since the latter's exact meaning to the user is not clear.

rmessner: Ok, so to be quite clear, i will use these variables in my default value in the script : 
spark_version
spark_distribution

rmessner:
Okay, i will use mirror instead of preferred_mirror to be consistent across flintrock

flintrock/config.yaml.template
nchammas: If you're talking about the config template, I think we just need a `{version}` variable. It seems unnecessary to say `spark_...` inside the Spark config.

As for the Hadoop distribution, we currently don't support specifying that, unless you are also intending to tackle #88.

nchammas:
#71 uses `download-source`, so I'd prefer that over `mirror`. I think it's clearer, though more verbose.

rmessner:
Okay, i'll use download-source instead of mirror.

For the variable names, it's okay for you as well ?

nchammas:
> For the variable names, it's okay for you as well ?

Not sure what you're referring to. I think the proposal laid out in #71 for Hadoop should give you a good template for what to name things.

If you're still not sure, it might be easier to just update the PR and I'll comment on the line items as necessary.

flintrock/config.yaml.template
```diff
@@ -1,6 +1,8 @@
 services:
   spark:
     version: 1.6.0
+    # distribution: # optional; default to '2.6'
```
nchammas: Style nitpick: Two spaces before the `#`; "defaults" and not "default"

flintrock/config.yaml.template
```diff
@@ -1,6 +1,8 @@
...
+    # download-source: # optional; default to 'https://s3.amazonaws.com/spark-related-packages/spark-${version}-bin-hadoop${distribution}.tgz'
```
nchammas: I prefer the variable substitution to be done in Python and not Bash, so the template variable should be `{version}` and not `${version}`. This is consistent with how Flintrock substitutes variables in the config templates, for example.

Same style nitpicks as above.

flintrock/scripts/install-spark.sh
```diff
@@ -2,22 +2,26 @@
 
 set -e
 
-spark_version="$1"
+version="$1"
 distribution="$2"
+download_source="$3"
+
+url=$(eval "echo \"$download_source\"")
```
nchammas: I think doing the variable substitution in Python should eliminate code smells like this one.

flintrock/config.yaml.template
nchammas: Hmm, can we leave out the ability to specify distribution for now? I'm not sure about how best to name this option (e.g. there are non-Hadoop distributions like CDH, but we are assuming Hadoop) and, more importantly, I haven't fully considered the implications of supporting user-specified distributions.

flintrock/flintrock.py
```diff
@@ -182,10 +182,17 @@ def cli(cli_context, config, provider):
...
+@click.option('--spark-distribution',
+              help="Hadoop distribution for Spark release to install.", default='2.6')
 @click.option('--spark-git-commit',
               help="Git commit to build Spark from. "
                    "Set to 'latest' to build Spark from the latest commit on the "
                    "repository's default branch.")
+@click.option('--spark-download-source',
+              help="HTTP source to download the spark binaries. "
+                   "Available variable : file, spark_version, distribution",
+              default="https://s3.amazonaws.com/spark-related-packages/spark-${version}-bin-hadoop${distribution}.tgz",
```
nchammas: Same comment about `{version}` vs. `${version}`.

BenFradet:
@rmessner are you still working on this?

Issue #14, opened 2015-10-07
Support launching clusters into private VPCs


nchammas:
Some users work in environments where they want to or have to launch clusters into VPCs with no public subnets.

This seems to be [a fairly common use-case](https://issues.apache.org/jira/issues/?jql=project%20%3D%20SPARK%20AND%20component%20%3D%20EC2%20AND%20text%20~%20%22private%20vpc%22%20ORDER%20BY%20updated%20DESC), so I think we  should support it if it does not add too much complexity.

A design goal for this feature should be to automatically do the appropriate setup whether the subnet we are launching into is public or private. That means, if possible, the user shouldn't have to specify anything.

I _think_ this is possible because we know what VPC we're launching into (either the user's default VPC, or an explicitly specified one), and we know what subnet we're launching into (either the VPC's default subnet, or an explicitly specified one). Flintrock should be able to query AWS for information about the subnet and figure out whether to use public or private addresses automatically.

MuksSharma:
Thanks @nchammas. This use case is becoming more common in enterprise setting. Any thoughts on when this feature might become available?

nchammas:
As with the typical open source project, there are no promises on when a feature might be delivered. The things that usually get worked on are critical bug fixes, fun stuff, and everything else, in that order. :smile: 

My current intent is to work next on adding support for resizing existing clusters and improving launch progress reporting. If I get some time at work to devote to Flintrock (e.g. for this issue), I may work on it after I get those 2 features out.

rmessner:
It is exactly our case ( and we don't even have hostname resolution, and that would need some hack for spark and hadoop ), I will work out something and submit it to you @nchammas.

My use case will be : 
* Downloading spark from a custom location ( thanks to #104 )
* Use private DNS names instead of public ones
* Define several origins ( multiple CIDR addresses to access the cluster for the security groups )

rmessner:
In case of the private VPC, I'm struggling with the ssh access when installing and configuring, because i'm outside the security group and it take my public IP to access an private IP so doesn't work. 

I'm thinking about 2 solutions : 
* Opening a security breach for the install and configuration step
* Bind the access_origins to the VPC ( means my private IP is within the range of the origins )

What do you think about it @nchammas ?

nchammas:
Hey @rmessner, I updated the description of the issue and commented on your PRs as well.

eshioji:
[...]

PR #105, opened 2016-04-06
Use a access_origins parameters for the security groups

### LINKED ISSUE #38: Give users an easy way to authorize new addresses to access a cluster
By default, Flintrock only authorizes the host it's running on to access launched clusters. So if you launch a cluster and then go somewhere else where your IP is gonna be different, you won't be able to login to the cluster anymore.

We should offer a way for users to authorize custom addresses on launch, and perhaps also a way for users to authorize their current address to access existing clusters.

rmessner:
This pull request is a partial/independant reply to the private VPC issue ( #14 )

Fixes #38.


nchammas:
@rmessner - Can you elaborate on the purpose of this PR?

If it does not stand alone as a valuable feature, but instead is part of the private VPC work, then could you collapse this PR into #108?

I'd like a single PR to cover adding private VPC support to Flintrock.

rmessner:
@nchammas - This feature alone allow user using Public Subnet to specify more than their own IP who is allowed to access spark pointed in this issue #38 .

But if you prefer a single PR anyway, i can easily merge it.

It's up to you

nchammas:
Ah, I see now. Sorry, I didn't realize at first that this was about #38. (I added a note to that effect to the original PR text so that the issue automatically gets closed when this PR is merged.)

I think this is fine as a separate PR. I will look at it more closely this week.

nchammas:
@rmessner - Now that you can associate specific security groups with a cluster on launch (#112), you can enable access to the cluster on any set of addresses via those security groups.

I prefer that approach since is it more general and since I think it can save us from having to add another option to the interface.

As a result, I'm going to close this PR. But please chime in if you disagree with my position or think I've missed something.

PR #108, opened 2016-04-08
Feature deploy in private vpc

rmessner:
This pull request combined with the "origins" pull request, allow to deploy in private VPC.

It offer an answer for #14


ryananguiano:
This seems to be working for me. I did get a warning about using a private subnet though. Is is still necessary to map_public_ip_on_launch on the subnet if you a you are using a private subnet.
- https://github.com/adotmob/flintrock/blob/feature-deploy-in-private-vpc/flintrock/ec2.py#L508

Also, when using private hostnames, I cannot resolve the internal ec2 hostnames with my connection. (Ex: ip-10-128-11-135.ec2.internal) Perhaps we could use IP address for internal hostname?
- https://github.com/adotmob/flintrock/blob/feature-deploy-in-private-vpc/flintrock/ec2.py#L84
- https://github.com/adotmob/flintrock/blob/feature-deploy-in-private-vpc/flintrock/ec2.py#L98

rmessner:
Thanks for your feedback.

Regarding the check about map_public_ip_on_launch, you are totally right, in case of private subnet, this is unnecessary to map public ip, i will change it on Monday.

About the private hostnames, it will look deeper in it to know where to make the relevant change.

nchammas:
@rmessner - Thanks for putting all this work into this issue. It's a tough one.

Is the `access-origins` option necessary? I know that [spark-ec2 supports launching into private VPCs](https://github.com/amplab/spark-ec2/blob/78c4a24b671eb433f9160e484e56a39894c9c106/spark_ec2.py#L320), but I don't remember that they needed anything other than a flag to say private vs. public IPs should be used.

More importantly, I was hoping to support private VPCs without requiring users to do anything. I don't know if this is even possible, but it's the approach I would want to explore first.

For example, we know what subnet we are launching into, either because it's the VPC's default subnet or because the user explicitly specified it. Is it possible to [query the subnet](https://boto3.readthedocs.org/en/latest/reference/services/ec2.html#subnet) and automatically determine whether to use public or private IPs?

If that's possible, then that's the way I would want to go. That way users don't have any extra options to deal with, and Flintrock just takes care of things automatically.

rmessner:
@nchammas - I think of a way to detect automatically if we are in public or private VPC, its would be if the instances have or not public hostname/ip. I will update my PR to take this into account.

nchammas:
Isn't it possible to determine this directly from the subnet by querying [`map_public_ip_on_launch`](https://boto3.readthedocs.org/en/latest/reference/services/ec2.html#EC2.Subnet.map_public_ip_on_launch)?

rmessner:
Exactly what i'm doing, but i didn't push yet.

nchammas:
Hey @rmessner I'm closing this PR due to inactivity and also since there is a similar proposal in #123 that takes the approach we discussed.

PR #109, opened 2016-04-08
Add the option to install hdfs from custom source

### LINKED ISSUE #71: Add option to download Hadoop from a custom URL

rmessner:
If a custom download source is provided, the system download directly from it.

Fixes #71.
Related to #104.

### DIFF
--- a/flintrock/scripts/download-hadoop.sh
+++ b/flintrock/scripts/download-hadoop.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+set -e
+
+version="$1"
+download_source="$2"
+
+url=$(eval "echo \"$download_source\"")
+file="${url##*/}"
+
+echo "Installing HDFS..."
+echo "  version: ${version}"
+echo "Final Hadoop URL: ${url}"
+
+# S3 is generally reliable, but sometimes when launching really large
+# clusters it can hiccup on us, in which case we'll need to retry the
+# download.
+set +e
+tries=1
+while true; do
+    curl --remote-name "${url}"
+    curl_ret=$?
+
+    if ((curl_ret == 0)); then
+        break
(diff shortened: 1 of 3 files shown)


flintrock/scripts/download-hadoop.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
nchammas: I'd rather reuse the existing Hadoop download script we have (`download-hadoop.py`) than add a new script alongside it. Can you do that?

Adding this script seems unnecessary.

nchammas:
I think a good approach to take here is have the default download source be `default`, which triggers the current behavior of querying Apache mirrors.

If the user overrides that, then we just skip over the mirror selection code within `download-hadoop.py` and go straight to the download.

BenFradet:
@rmessner are you still working on this?  
If not, do you mind if I take over?

PR #114, opened 2016-05-03
Add security rule: open port 6066 from client

dm-tran:
* open port 6066 from client, to allow '--deploy-mode cluster' when submitting jobs

I tested this PR by submitting a spark job from a remote client (a host that does not belong to the spark cluster) : `spark-submit --deploy-mode cluster --master spark://host:6066`

### DIFF
--- a/flintrock/ec2.py
+++ b/flintrock/ec2.py
@@ -374,6 +374,7 @@ def get_or_create_ec2_security_groups(
             to_port=7077,
             cidr_ip=flintrock_client_cidr,
             src_group=None),
+        # Spark REST Server
         SecurityGroupRule(
             ip_protocol='tcp',
             from_port=6066,
--- a/flintrock/ec2.py
+++ b/flintrock/ec2.py
@@ -373,6 +373,12 @@ def get_or_create_ec2_security_groups(
             from_port=7077,
             to_port=7077,
             cidr_ip=flintrock_client_cidr,
+            src_group=None),
+        SecurityGroupRule(
+            ip_protocol='tcp',
+            from_port=6066,
+            to_port=6066,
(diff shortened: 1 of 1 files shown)


nchammas:
Thanks for the PR @dm-tran.

How does this relate to #103? I thought port 7077 was the Spark master port.

nchammas:
I couldn't find mention of port 6066 anywhere in the docs, but I did find [this line](https://github.com/apache/spark/blob/947b9020b0d621bc97661a0a056297e6889936d3/core/src/main/scala/org/apache/spark/deploy/master/Master.scala#L154) which seems to suggest the port is for a REST server.

Is that required for `spark-submit --deploy-mode cluster` to work from a remote client?

dm-tran:
Port 7077 is used with '--deploy-mode client' (which is the default deploy-mode).
Port 6066 is used with '--deploy-mode cluster' and Spark's REST API (http://arturmkrtchyan.com/apache-spark-hidden-rest-api).

The main page of Spark UI (http://master-host:8080/) lists both ports : 
```
URL: spark://host:7077
REST URL: spark://host:6066 (cluster mode)
```

nchammas:
Interesting. Thanks for the link!

flintrock/ec2.py
```diff
@@ -373,6 +373,12 @@ def get_or_create_ec2_security_groups(
             from_port=7077,
             to_port=7077,
             cidr_ip=flintrock_client_cidr,
+            src_group=None),
+        SecurityGroupRule(
```
nchammas: Could you add a brief comment here so that down the line we remember what this port is for?

```python
# Spark REST Server
SecurityGroupRule(
    ...
```

PR #118, opened 2016-05-21
Add option to download Hadoop from a custom URL

### LINKED ISSUE #71: Add option to download Hadoop from a custom URL

BenFradet:
* add an option to specify your own mirror to download hadoop from

I did some manual tests:
  - no `download_source`
  - working `download_source`
  - faulty `download_source`

Fixes #71.
Supersedes / Fixes #109.

### DIFF
--- a/flintrock/config.yaml.template
+++ b/flintrock/config.yaml.template
@@ -5,6 +5,9 @@ services:
     # git-repository:  # optional; defaults to https://github.com/apache/spark
   hdfs:
     version: 2.7.2
+    # download-source: # optional; defaults to
+    # http://www.apache.org/dyn/closer.lua/hadoop/common/hadoop-{v}/hadoop-{v}.tar.gz?as_json
+    # and must contain a {v} template corresponding to the version
 
 provider: ec2
 
--- a/flintrock/config.yaml.template
+++ b/flintrock/config.yaml.template
@@ -5,9 +5,9 @@ services:
     # git-repository:  # optional; defaults to https://github.com/apache/spark
   hdfs:
     version: 2.7.2
-    # download-source: # optional; defaults to
-    # http://www.apache.org/dyn/closer.lua/hadoop/common/hadoop-{v}/hadoop-{v}.tar.gz?as_json
-    # and must contain a {v} template corresponding to the version
+    # optional; defaults to download from a dynamically selected Apache mirror
+    # must contain a {v} template corresponding to the version; must be a .tar.gz file
+    # download-source: "https://www.example.com/files/hadoop/{v}/hadoop-{v}.tar.gz"
 
 provider: ec2
 
--- a/flintrock/flintrock.py
+++ b/flintrock/flintrock.py
@@ -179,6 +179,10 @@ def cli(cli_context, config, provider):
 @click.option('--num-slaves', type=int, required=True)
 @click.option('--install-hdfs/--no-install-hdfs', default=False)
 @click.option('--hdfs-version')
+@click.option('--hdfs-download-source',
+              help="Custom URL to download hadoop from.",
+              default='http://www.apache.org/dyn/closer.lua/hadoop/common/hadoop-{v}/hadoop-{v}.tar.gz?as_json',
+              show_default=True)
 @click.option('--install-spark/--no-install-spark', default=True)
 @click.option('--spark-version',
               help="Spark release version to install.")
@@ -218,6 +222,7 @@ def launch(
         num_slaves,
         install_hdfs,
         hdfs_version,
+        hdfs_download_source,
         install_spark,
         spark_version,
         spark_git_commit,
@@ -276,13 +281,11 @@ def launch(
     #      https://github.com/mitchellh/packer/issues/1935#issuecomment-111235752
     option_requires(
         option='--ec2-vpc-id',
-        requires_all=[
-            '--ec2-subnet-id'
-        ],
+        requires_all=['--ec2-subnet-id'],
         scope=locals())
 
     if install_hdfs:
-        hdfs = HDFS(version=hdfs_version)
+        hdfs = HDFS(version=hdfs_version, download_source=hdfs_download_source)
         services += [hdfs]
     if install_spark:
         if spark_version:
--- a/flintrock/flintrock.py
+++ b/flintrock/flintrock.py
@@ -180,7 +180,7 @@ def cli(cli_context, config, provider):
 @click.option('--install-hdfs/--no-install-hdfs', default=False)
 @click.option('--hdfs-version')
 @click.option('--hdfs-download-source',
-              help="Custom URL to download hadoop from.",
+              help="URL to download Hadoop from.",
               default='http://www.apache.org/dyn/closer.lua/hadoop/common/hadoop-{v}/hadoop-{v}.tar.gz?as_json',
               show_default=True)
 @click.option('--install-spark/--no-install-spark', default=True)
--- a/flintrock/scripts/download-hadoop.py
+++ b/flintrock/scripts/download-hadoop.py
@@ -1,11 +1,10 @@
 """
-Download Hadoop from the best available Apache mirror.
+Download Hadoop from the best available Apache mirror or a custom location.
 """
 
 from __future__ import print_function
 
 import json
-import os
 import sys
 import subprocess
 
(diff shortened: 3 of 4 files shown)


BenFradet:
@nchammas as a spin off of this we could also support `latest` as hadoop and spark version, what do you think?

flintrock/services.py
```diff
@@ -129,14 +130,14 @@ def install(
...
                 set -e
 
-                python /tmp/download-hadoop.py "{version}"
+                python /tmp/download-hadoop.py "{version}" "{download_source}"
 
                 mkdir "hadoop"
                 mkdir "hadoop/conf"
 
                 tar xzf "hadoop-{version}.tar.gz" -C "hadoop" --strip-components=1
                 rm "hadoop-{version}.tar.gz"
```
BenFradet: I'm a bit concerned about this since the file name could be not `hadoop-{version}.tar.gz` or, alternatively, we could impose the file name to be `hadoop-{version}.tar.gz` when `urlretrieve`ing, what do you think?

nchammas:
> as a spin off of this we could also support latest as hadoop and spark version, what do you think?

That's a good idea and is consistent with the type of user experience we want to offer. I worry though that this will require manually maintaining the latest version, which I would be against.

Perhaps we could query the [releases on GitHub](https://github.com/apache/spark/releases), excluding preview and release candidates, to get the latest version. We'd have to be careful though, as the latest release by date is not necessarily the latest version. For example, Spark 1.6.2 will probably be released after Spark 2.0.

Anyway, we don't need to do this as part of this PR, but if you want to open a separate issue to discuss it, go ahead. I think it's a good idea.

flintrock/services.py
nchammas: > we could impose the file name to be `hadoop-{version}.tar.gz` when `urlretrieve`ing

Sounds good to me.

flintrock/scripts/download-hadoop.py
```diff
@@ -18,16 +18,27 @@
 
 if __name__ == '__main__':
     hadoop_version = sys.argv[1]
-    apache_mirror_url = "http://www.apache.org/dyn/closer.lua/hadoop/common/hadoop-{v}/hadoop-{v}.tar.gz?as_json".format(v=hadoop_version)
+    custom_mirror_url = sys.argv[2]
```
nchammas: Hmm... It looks like `custom_mirror_url` will always be populated, even if the download setting is left to the default. No? I'm looking over at [this line](https://github.com/nchammas/flintrock/pull/118/files#diff-2866cac17a48e5c58161518d728d0f8fR133).

flintrock/scripts/download-hadoop.py
```diff
@@ -18,16 +18,27 @@
...
+        mirror_url = "http://www.apache.org/dyn/closer.lua/hadoop/common/hadoop-{v}/hadoop-{v}.tar.gz?as_json"\
+            .format(v=hadoop_version)
 
     tries = 0
     while tries < 3:
-        mirror_info = json.loads(urlopen(apache_mirror_url).read().decode('utf-8'))
-        file_url = mirror_info['preferred'] + mirror_info['path_info']
-        file_name = os.path.basename(mirror_info['path_info'])
-        print('Downloading file at:', file_url)
+        if custom_mirror_url:
```
nchammas: Can we collapse these two branches somehow? I think parts like the `print()` and `urlretrieve()` should be written once in a common block after the `if`/`else` branch.

Same goes for the `if`/`else` branch just above--I'd prefer it if we could have just one branch in this whole file that does different stuff based on whether the URL is custom or not, and the rest of the code should be the same.

flintrock/scripts/download-hadoop.py
BenFradet: yes indeed

flintrock/flintrock.py
```diff
@@ -179,6 +179,10 @@ def cli(cli_context, config, provider):
 @click.option('--num-slaves', type=int, required=True)
 @click.option('--install-hdfs/--no-install-hdfs', default=False)
 @click.option('--hdfs-version')
+@click.option('--hdfs-download-source',
+              help="Custom URL to download hadoop from.",
```
nchammas: Nitpick: This should just be "URL to download Hadoop from." since even the default URL is specified here. Capitalize the "Hadoop" too, please. :)

flintrock/scripts/download-hadoop.py
BenFradet: I'll look into it

flintrock/config.yaml.template
```diff
@@ -5,6 +5,9 @@ services:
     # git-repository:  # optional; defaults to https://github.com/apache/spark
   hdfs:
     version: 2.7.1
+    # download-source: # optional; defaults to
+    # http://www.apache.org/dyn/closer.lua/hadoop/common/hadoop-{v}/hadoop-{v}.tar.gz?as_json
```
nchammas: Since they are lengthy, let's move all the comment lines for this option to right above the `# download-source: ` line.

nchammas:
@BenFradet - Thank you for this PR! I left some minor comments about style and structure, but this looks very good to me.

BenFradet:
Thanks for review, will make the necessary changes.

flintrock/config.yaml.template
BenFradet: Do you mean:
```yaml
# optional; defaults to
# http://www.apache.org/dyn/closer.lua/hadoop/common/hadoop-{v}/hadoop-{v}.tar.gz?as_json
# and must contain a {v} template corresponding to the version
# download-source:
```
?

flintrock/scripts/download-hadoop.py
```diff
@@ -18,16 +18,27 @@
...
+            print("Downloading file at:", mirror_url)
+            file_name = os.path.basename(mirror_url)
+            file_path, _ = urlretrieve(url=mirror_url, filename=file_name)
+        else:
+            mirror_info = json.loads(urlopen(mirror_url).read().decode('utf-8'))
```
BenFradet: Since retries are only based on archive integrity, should we also retry the json?

flintrock/config.yaml.template
nchammas: Yep.

flintrock/scripts/download-hadoop.py
nchammas: If you mean re-querying `http://www.apache.org/dyn/closer.lua/`, then yes. Sometimes an Apache mirror will give us a bad file, so we need to re-query `closer.lua` to get a different mirror.

flintrock/scripts/download-hadoop.py
```diff
@@ -18,16 +17,22 @@
...
     while tries < 3:
-        mirror_info = json.loads(urlopen(apache_mirror_url).read().decode('utf-8'))
-        file_url = mirror_info['preferred'] + mirror_info['path_info']
-        file_name = os.path.basename(mirror_info['path_info'])
-        print('Downloading file at:', file_url)
+        if mirror_url == default_mirror_url:
+            mirror_info = json.loads(urlopen(mirror_url).read().decode('utf-8'))
+            file_url = mirror_info['preferred'] + mirror_info['path_info']
+        else:
+            file_url = mirror_url
```
BenFradet: I tried to streamline this as much as possible, what do you think @nchammas ?

nchammas: Much better. :)

At some point in the future I may use `argparse` in this script and try to get rid of the duplicate default URL in the Click command spec. But that's for future work.

flintrock/config.yaml.template
```diff
@@ -5,6 +5,10 @@ services:
     # git-repository:  # optional; defaults to https://github.com/apache/spark
   hdfs:
     version: 2.7.1
+    # optional; defaults to
+    # http://www.apache.org/dyn/closer.lua/hadoop/common/hadoop-{v}/hadoop-{v}.tar.gz?as_json
+    # and must contain a {v} template corresponding to the version
+    # download-source:
```
nchammas: Hmm, I just realized that another requirement we have is that the file be a `.tar.gz`. Could we update the comment to the following?

```
    # optional; defaults to download from a dynamically selected Apache mirror
    # must contain a {v} template corresponding to the version; must be a .tar.gz file
    # download-source: "https://www.example.com/files/hadoop/{v}/hadoop-{v}.tar.gz"
```

nchammas:
Tested this out; works as expected.

Excellent work @BenFradet. I had one last minor comment; once that's addressed I'll merge this in.

nchammas:
Oops, I made a small change that introduced a merge conflict here. Sorry @BenFradet!

You'll need to rebase against master and resolve the conflict. Feel free to ping me if you need help doing that.

BenFradet:
Yeah np, I'll do that later today.

nchammas:
Great work @BenFradet. I've merged this in. Thanks!

PR #123, opened 2016-05-26
Add support for launching in private vpc

### LINKED ISSUE #14: Support launching clusters into private VPCs

jperezdiaz:
*  Queries [map_public_ip_on_launch](https://boto3.readthedocs.io/en/latest/reference/services/ec2.html#EC2.Subnet.map_public_ip_on_launch) to figure out if the subnet auto-assigns public ips

Fixes #14 .

### DIFF
--- a/flintrock/core.py
+++ b/flintrock/core.py
@@ -513,8 +513,6 @@ def provision_node(
                 fullname=`hostname`.ec2.internal
 
                 echo "{h} $fullname $(hostname)" |sudo tee -a /etc/hosts
-
-                set +e
                 """.format(h=host))
 
         for service in services:
--- a/flintrock/core.py
+++ b/flintrock/core.py
@@ -93,6 +93,16 @@ class FlintrockCluster:
         """
         raise NotImplementedError
 
+    @property
+    def subnet_is_private(self) -> bool:
+        """
+        A boolean set to true if the subnet is private..
+
+        Providers must override this property since it is typically derived from
+        an underlying object, like an EC2 instance.
+        """
+        raise NotImplementedError
+
     def destroy_check(self):
         """
         Check that the cluster is in a state in which it can be destroyed.
@@ -504,16 +514,17 @@ def provision_node(
                     sudo sh -c "echo export JAVA_HOME=/usr/lib/jvm/jre >> /etc/environment"
                     source /etc/environment
                 """)
-        print("[{h}] Configuring hostname...".format(h=host))
-        ssh_check_output(
-            client=client,
-            command="""
-                set -e
+        if cluster.subnet_is_private:
+            print("[{h}] Configuring hostname...".format(h=host))
+            ssh_check_output(
+                client=client,
+                command="""
+                    set -e
 
-                fullname=`hostname`.ec2.internal
+                    fullname=`hostname`.ec2.internal
 
-                echo "{h} $fullname $(hostname)" |sudo tee -a /etc/hosts
-                """.format(h=host))
+                    echo "{h} $fullname $(hostname)" |sudo tee -a /etc/hosts
+                    """.format(h=host))
 
         for service in services:
             service.install(
--- a/flintrock/core.py
+++ b/flintrock/core.py
@@ -504,6 +504,18 @@ def provision_node(
                     sudo sh -c "echo export JAVA_HOME=/usr/lib/jvm/jre >> /etc/environment"
                     source /etc/environment
                 """)
+        print("[{h}] Configuring hostname...".format(h=host))
+        ssh_check_output(
+                client=client,
+                command="""
+                    set -e
+                    
+                    fullname=`hostname`.ec2.internal
+
+                    echo "{h} $fullname $(hostname)" |sudo tee -a /etc/hosts
+
+                    set +e
+                    """.format(h=host))
 
         for service in services:
             service.install(
--- a/flintrock/core.py
+++ b/flintrock/core.py
@@ -506,16 +506,16 @@ def provision_node(
                 """)
         print("[{h}] Configuring hostname...".format(h=host))
         ssh_check_output(
-                client=client,
-                command="""
-                    set -e
-                    
-                    fullname=`hostname`.ec2.internal
+            client=client,
+            command="""
+                set -e
 
-                    echo "{h} $fullname $(hostname)" |sudo tee -a /etc/hosts
+                fullname=`hostname`.ec2.internal
 
-                    set +e
-                    """.format(h=host))
+                echo "{h} $fullname $(hostname)" |sudo tee -a /etc/hosts
+
+                set +e
+                """.format(h=host))
 
         for service in services:
             service.install(
--- a/flintrock/ec2.py
+++ b/flintrock/ec2.py
@@ -58,7 +58,6 @@ class EC2Cluster(FlintrockCluster):
             vpc_id: str,
             master_instance: 'boto3.resources.factory.ec2.Instance',
             slave_instances: "List[boto3.resources.factory.ec2.Instance]",
-            use_private_vpc,
             *args,
             **kwargs):
         super().__init__(*args, **kwargs)
@@ -66,7 +65,6 @@ class EC2Cluster(FlintrockCluster):
         self.vpc_id = vpc_id
         self.master_instance = master_instance
         self.slave_instances = slave_instances
-        self.use_private_vpc = use_private_vpc
 
     @property
     def instances(self):
@@ -74,31 +72,36 @@ class EC2Cluster(FlintrockCluster):
 
     @property
     def master_ip(self):
-        if self.use_private_vpc:
+        if self.subnet_is_private:
             return self.master_instance.private_ip_address
         else:
             return self.master_instance.public_ip_address
 
     @property
     def master_host(self):
-        if self.use_private_vpc:
+        if self.subnet_is_private:
             return self.master_instance.private_dns_name
         else:
             return self.master_instance.public_dns_name
 
     @property
     def slave_ips(self):
-        if self.use_private_vpc:
+        if self.subnet_is_private:
             return [i.private_ip_address for i in self.slave_instances]
         else:
             return [i.public_ip_address for i in self.slave_instances]
 
     @property
     def slave_hosts(self):
-        if self.use_private_vpc:
+        if self.subnet_is_private:
             return [i.private_dns_name for i in self.slave_instances]
         else:
             return [i.public_dns_name for i in self.slave_instances]
+
+    @property
+    def subnet_is_private(self):
+        ec2 = boto3.resource(service_name='ec2', region_name=self.region)
+        return not ec2.Subnet(self.master_instance.subnet_id).map_public_ip_on_launch
 
     @property
     def state(self):
@@ -649,8 +652,7 @@ def launch(
             vpc_id=vpc_id,
             ssh_key_pair=generate_ssh_key_pair(),
             master_instance=master_instance,
-            slave_instances=slave_instances,
-            use_private_vpc=use_private_vpc)
+            slave_instances=slave_instances)
 
         cluster.wait_for_state('running')
 
@@ -802,15 +804,11 @@ def _compose_cluster(*, name: str, region: str, vpc_id: str, instances: list) ->
     """
     (master_instance, slave_instances) = _get_cluster_master_slaves(instances)
 
-    ec2 = boto3.resource(service_name='ec2', region_name=region)
-    use_private_vpc = not ec2.Subnet(master_instance.subnet_id).map_public_ip_on_launch
-
     cluster = EC2Cluster(
         name=name,
         region=region,
         vpc_id=vpc_id,
         master_instance=master_instance,
-        slave_instances=slave_instances,
-        use_private_vpc=use_private_vpc)
+        slave_instances=slave_instances)
 
     return cluster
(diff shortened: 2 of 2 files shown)


BenFradet:
You can run `py.test tests/test_static.py` to check style issues as detailed in the [test guide](https://github.com/nchammas/flintrock/tree/master/test).

flintrock/core.py
```diff
@@ -504,6 +504,18 @@ def provision_node(
                     sudo sh -c "echo export JAVA_HOME=/usr/lib/jvm/jre >> /etc/environment"
                     source /etc/environment
                 """)
+        print("[{h}] Configuring hostname...".format(h=host))
+        ssh_check_output(
+            client=client,
+            command="""
+                set -e
+
+                fullname=`hostname`.ec2.internal
```
nchammas: `core.py` should not have any logic that's specific to a provider like EC2. If someone added GCE support to Flintrock tomorrow, we'd want to be able to reuse all the logic in `core.py` mostly as-is.

flintrock/core.py
```diff
@@ -504,6 +504,18 @@ def provision_node(
...
+
+                echo "{h} $fullname $(hostname)" |sudo tee -a /etc/hosts
```
nchammas: Why do we need to add a new `hosts` entry? Does the cluster fail to launch without this?

flintrock/core.py
```diff
@@ -504,6 +504,18 @@ def provision_node(
...
+
+                set +e
```
nchammas: I don't think you need to do this. I believe settings like `-e` get reset between SSH sessions.

flintrock/ec2.py
```diff
@@ -57,33 +58,47 @@ def __init__(
             vpc_id: str,
             master_instance: 'boto3.resources.factory.ec2.Instance',
             slave_instances: "List[boto3.resources.factory.ec2.Instance]",
+            use_private_vpc,
```
nchammas: Does this need to be part of the object signature? It seems that whether or not the VPC is private is simply a derived property of `vpc_id`.

flintrock/core.py
jorgito1167: That is a good point. Lets figure out if it is needed at all and I'll move or remove. The start-master.sh and start-slave.sh scripts were failing because the nodes could not find their ip addresses using their hostname. I added themselves to the hosts file.

flintrock/ec2.py
```diff
@@ -274,20 +289,19 @@ def check_network_config(*, region_name: str, vpc_id: str, subnet_id: str):
...
             .format(v=vpc_id)
         )
-    if not ec2.Subnet(subnet_id).map_public_ip_on_launch:
-        raise ConfigurationNotSupported(
-            "{s} does not auto-assign public IP addresses. "
-            "Flintrock requires public IP addresses.\n"
-            "See: https://github.com/nchammas/flintrock/issues/14"
-            .format(s=subnet_id)
-        )
+    use_private_vpc = not ec2.Subnet(subnet_id).map_public_ip_on_launch
```
nchammas: As noted earlier, I think this belongs as a property of `EC2Cluster`. I think all we need to do here is remove the `ConfigurationNotSupported` for private VPCs, as you've done.

nchammas:
Thanks for taking this on @jorgito1167!

This looks like a good start. I think this PR can be made simpler and better by making `vpc_is_private` a `@property` of the `EC2Cluster` class. That will eliminate the need for many of the little changes that have been made to helper methods in `ec2.py`.

flintrock/core.py
nchammas: Ah interesting. So they were trying the correct hostname (the private one) and it wasn't resolving?

jorgito1167: Yes. The hostname they get is ip-x-x-x-x and the private DNS is ip-x-x-x-x.ec2.internal. You can't ping hostname but you can ping the private DNS.

flintrock/ec2.py
```diff
@@ -334,9 +348,12 @@ def get_or_create_ec2_security_groups(
             VpcId=vpc_id)
 
     # Rules for the client interacting with the cluster.
-    flintrock_client_ip = (
-        urllib.request.urlopen('http://checkip.amazonaws.com/')
-        .read().decode('utf-8').strip())
+    if use_private_vpc:
```
jorgito1167: If the host running Flintrock does not have a public ip, the past method for getting the ip address won't return the private ip. The current solution assumes that the user will always are launch a cluster into a private subnet from a host with no public ip. Is there a better way of handling this?

nchammas: Good question. I think whether the Flintrock client has a private IP is distinct from whether the cluster is in a private VPC (though often the two will go together). So we need a way of determining what IP to use for the Flintrock client that doesn't depend on the cluster.

Perhaps we can simply query `checkip` first, and if that fails fallback to `gethostbyname()`?

flintrock/ec2.py
jorgito1167: I'm not sure that the `self.vpc_id` is enough to determine if the subnet that the cluster is in auto-assigns public ips. However, I can get the subnet id from the master instance and then query the subnet to get the desired result. What do you think?

flintrock/ec2.py
jorgito1167: The problem is that it does not fail. It simply returns a public IP as seen by the aws checkip server.

flintrock/ec2.py
nchammas: Perhaps then we need to add `subnet_id` as a cluster attribute and use that. Currently, the user can explicitly provide a subnet to use, or Flintrock will assume it can be determined automatically from the VPC, because a VPC can have a default subnet.

I don't think we need to query the master instance.

flintrock/ec2.py
nchammas: Ah, right, and that won't be the IP that the cluster sees when Flintrock tries to connect, if both the client and the cluster are together on a private subnet.

Perhaps then we should always authorize both addresses, public and private?

flintrock/ec2.py
jorgito1167: The plan is then to add an attribute that keeps the subnet_id. Do we include a subnet_is_private method that queries the subnet to determine if it private? In the case we make this change, get_cluster, get_clusters and _compose_cluster also need the subnet_id.

flintrock/ec2.py
jorgito1167: That might be dangerous since the IP that results from querying the checkip server is common to all of the computers under the NAT instance. I'm not sure if it will allow access from other computers.

flintrock/ec2.py
nchammas: > Do we include a subnet_is_private method that queries the subnet to determine if it private?

I think `subnet_is_private` should be a `@property`.

> In the case we make this change, get_cluster, get_clusters and _compose_cluster also need the subnet_id.

Do you mean as an input parameter? Ahhhh... I see what you were saying earlier now.

I think the signatures of these methods should stay the same (we don't need the subnet to locate a cluster), but when they return a cluster object it should have the `subnet_id` defined.

We will need to query an instance (e.g. the master) to get that information, like you were suggesting earlier. My apologies.

flintrock/ec2.py
nchammas: That seems fairly innocuous to me, since those other machines would be under your account. Or are you saying machines from different AWS accounts may share the same public IP address?

jorgito1167: I'm not sure to be honest. I would like to think that it doesn't happen but it would be great to come up with a solution that only authorizes both IPs. For the purposes of this PR we can just go with authorizing both. Later we can include an option to authorize only a specific security group or list or ip addresses that the user provides.

nchammas: I've reviewed the relevant docs here:
* http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_Introduction.html
* http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/vpc-nat.html

It sounds like your VPC has an internet gateway attached, otherwise the call to `checkip` would fail, or perhaps would return a private address.

Even with the gateway attached, from my reading of the docs, it sounds like only instances from the same VPC can ever get the same public IP address. So I don't think it's an issue to authorize both addresses.

jperezdiaz: Sounds good. I'll add a `try except` for handling the case when it fails.

nchammas:
Hey @jperezdiaz (did you recently change your username?), I took a quick look through your latest changes and things generally look good.

There are still some open items, though:
* `core.py` still includes some logic that is specific to EC2. We need to factor that out, or possibly do away with it entirely. Do you have any suggestions in that regard? Is `instance.private_dns_name` not giving us something usable "out-of-the-box"?
* I just tried launching a cluster off of this PR using my current config with a public VPC and it failed. Well, the launch technically succeeded but Spark couldn't come up. It looks like the master had trouble binding to an address. Does this work for you?

jperezdiaz:
Yes, I changed my username recently. Sorry for the confusion.

* It seems like public VPC did not have problems with the hostname. We can condition the  `/etc/hosts` script on the subnet being private. I also like the idea of using getting the instance private DNS to replace the hardcoded `.ec2.internal`. 

* How can I test if the master successfully bound to an address. I'm pretty sure I'm having a similar problem even with the private VPC.

nchammas:
> How can I test if the master successfully bound to an address? I'm pretty sure I'm having a similar problem even with the private VPC.

When you launch a cluster and the master fails to start properly, the Spark health check should show 0 workers. Then, if you login to the cluster and start a shell (either `spark-shell` or `pyspark`), you'll also get an error. It should be pretty obvious.

jperezdiaz:
It is non-trivial for me to test in a public vpc. However, it seems to launch fine in the private vpc. The health check reports 1 worker and I'm able to start pyspark. Could the problem be due to the change in the `/etc/hosts` file? In that case, we can just condition the change on the subnet being private.

On an unrelated note, the problem I see is that when I start pyspark I get the following message for the Spark UI:

```
16/05/31 17:55:17 INFO SparkUI: Started SparkUI at http://<?xml version="1.0" encoding="iso-8859-1"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
         "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
  <title>404 - Not Found</title>
 </head>
 <body>
  <h1>404 - Not Found</h1>
 </body>
</html>:4040
```

nchammas:
> Could the problem be due to the change in the /etc/hosts file? In that case, we can just condition the change on the subnet being private.

I'm not sure, but I am suspicious of that change.

> On an unrelated note, the problem I see is that when I start pyspark I get the following message for the Spark UI:

You're seeing this when you launch in a private VPC?

jperezdiaz:
* Is there a good way of figuring out the private DNS of the instances from inside the `provision_node` function? The instances attribute is specific to the EC2 cluster so it should not be used. I could get the `slave_ips` list, get the index of the current ip and then use it to index the private DNS list. Otherwise, I can create a method in for the `FlintrockCluster` object that does this for us. 

* Yes, the Spark UI problem happens when I launch in a private VPC.

nchammas:
> Is there a good way of figuring out the private DNS of the instances from inside the provision_node function?

I think `provision_node()` and the rest of the code in `core.py` should ideally just know about IP addresses, and not distinguish whether they are public or private.

> Yes, the Spark UI problem happens when I launch in a private VPC.

Does Spark work otherwise?

I'm trying to replicate your setup so I can help you test this. Can you lay out the VPC, subnet, routing table, etc. you have and how they are configured so I can setup a parallel environment?

Ideally we should capture this setup as code and use it in an acceptance test (e.g. setup private VPC, test launch/Spark/HDFS, tear down cluster and VPC), but that's probably a bit much for now. We can add that in later, unless you feel like having a go at it now.

jperezdiaz:
I like the idea of setting up the test. I'm really busy this week but I'll try to implement it once I get some time.

nchammas:
Hey @jperezdiaz are you interested in updating this PR?

Looking through the history, it looks like we agreed on the basic approach you took here, but there were 2 unresolved issues:

1. EC2-specific logic was added to `core.py`. `core.py` should be completely provider-agnostic.
2. You were experiencing some issues with the launch and/or Spark UI.

If you aren't planning to update it anytime soon, we can close the PR for now and you can revisit it when you are ready.

nchammas:
Closing this PR. Feel free to open a new one if you are interested in continuing this work!

PR #124, opened 2016-05-26
Correct default values for install-{spark,hdfs} in the config template

BenFradet:
I noticed that the values in the config template weren't reflecting the default values for install-{spark,hdfs}, this pr aims to fix that.

### DIFF
--- a/flintrock/config.yaml.template
+++ b/flintrock/config.yaml.template
@@ -32,5 +32,5 @@ providers:
 
 launch:
   num-slaves: 1
-  # install-hdfs: True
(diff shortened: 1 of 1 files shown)


nchammas:
Ah, this was intentional. The idea is that we want to default to install Spark but not HDFS, and if the user wants to change that they can start by simply uncommenting these lines in the config file.

I haven't been following this pattern consistently throughout the config file, but does it make sense?

BenFradet:
Seems a bit weird to me but I get it yes.

PR #125, opened 2016-06-22
Add option to download Spark from a custom URL

### LINKED ISSUE #88: Cannot install package pre-built for Hadoop 2.4

### LINKED ISSUE #101: Add option to download Spark from a custom URL

BenFradet:
This PR adds a `download_source` to the Spark service as has been done in #118.

I created clusters with both with and without a `download_source` to test the new feature.

Fixes #101, fixes #88.
Closes #104.

### DIFF
--- a/flintrock/config.yaml.template
+++ b/flintrock/config.yaml.template
@@ -3,13 +3,16 @@ services:
     version: 1.6.1
     # git-commit: latest  # if not 'latest', provide a full commit SHA; e.g. d6dc12ef0146ae409834c78737c116050961f350
     # git-repository:  # optional; defaults to https://github.com/apache/spark
-    # optional; defaults to download from a dynamically selected Apache mirror
-    # must contain a {v} template corresponding to the version; must be a .tar.gz file
+    # optional; defaults to download from from the official Spark S3 bucket
+    #   - must contain a {v} template corresponding to the version
+    #   - Spark must be pre-built
+    #   - must be a tar.gz file
     # download-source: "https://www.example.com/files/spark/{v}/spark-{v}.tar.gz"
   hdfs:
     version: 2.7.2
     # optional; defaults to download from a dynamically selected Apache mirror
-    # must contain a {v} template corresponding to the version; must be a .tar.gz file
+    #   - must contain a {v} template corresponding to the version
+    #   - must be a .tar.gz file
     # download-source: "https://www.example.com/files/hadoop/{v}/hadoop-{v}.tar.gz"
 
 provider: ec2
--- a/flintrock/config.yaml.template
+++ b/flintrock/config.yaml.template
@@ -3,6 +3,9 @@ services:
     version: 1.6.1
     # git-commit: latest  # if not 'latest', provide a full commit SHA; e.g. d6dc12ef0146ae409834c78737c116050961f350
     # git-repository:  # optional; defaults to https://github.com/apache/spark
+    # optional; defaults to download from a dynamically selected Apache mirror
+    # must contain a {v} template corresponding to the version; must be a .tar.gz file
+    # download-source: "https://www.example.com/files/spark/{v}/spark-{v}.tar.gz"
   hdfs:
     version: 2.7.2
     # optional; defaults to download from a dynamically selected Apache mirror
--- a/flintrock/flintrock.py
+++ b/flintrock/flintrock.py
@@ -194,6 +194,10 @@ def cli(cli_context, config, provider):
               help="Git repository to clone Spark from.",
               default='https://github.com/apache/spark',
               show_default=True)
+@click.option('--spark-download-source',
+              help="URL to download Spark from.",
+              default='https://s3.amazonaws.com/spark-related-packages/spark-{v}-bin-hadoop2.6.tgz',
+              show_default=True)
 @click.option('--assume-yes/--no-assume-yes', default=False)
 @click.option('--ec2-key-name')
 @click.option('--ec2-identity-file',
@@ -227,6 +231,7 @@ def launch(
         spark_version,
         spark_git_commit,
         spark_git_repository,
+        spark_download_source,
         assume_yes,
         ec2_key_name,
         ec2_identity_file,
@@ -289,7 +294,7 @@ def launch(
         services += [hdfs]
     if install_spark:
         if spark_version:
-            spark = Spark(version=spark_version)
+            spark = Spark(version=spark_version, download_source=spark_download_source)
         elif spark_git_commit:
             print(
                 "Warning: Building Spark takes a long time. "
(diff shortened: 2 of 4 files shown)


flintrock/services.py
```diff
@@ -202,28 +202,29 @@ def health_check(self, master_host: str):
...
+            'download_source': download_source,
             'git_commit': git_commit,
             'git_repository': git_repository}
 
     def install(
             self,
             ssh_client: paramiko.client.SSHClient,
             cluster: FlintrockCluster):
-        # TODO: Allow users to specify the Spark "distribution". (?)
-        distribution = 'hadoop2.6'
```
BenFradet: As a follow up we could support a `{d}` template in `download_source` as is done with the version with `{v}`.

flintrock/scripts/install-spark.sh
```diff
@@ -2,22 +2,20 @@
```
BenFradet: Out of curiosity, why is the `download-hadoop` script written in python, whereas this one is in bash?

flintrock/config.yaml.template
```diff
@@ -3,6 +3,9 @@ services:
     version: 1.6.1
     # git-commit: latest  # if not 'latest', provide a full commit SHA; e.g. d6dc12ef0146ae409834c78737c116050961f350
     # git-repository:  # optional; defaults to https://github.com/apache/spark
+    # optional; defaults to download from a dynamically selected Apache mirror
+    # must contain a {v} template corresponding to the version; must be a .tar.gz file
+    # download-source: "https://www.example.com/files/spark/{v}/spark-{v}.tar.gz"
```
BenFradet: Maybe we should as well mention that the `download_source` has to be prebuilt?

flintrock/services.py
nchammas: Agreed, and that would address #88, though it seems like with this PR you can already choose your distribution at will, right?

flintrock/scripts/install-spark.sh
nchammas: This is an unfortunately inconsistency, and eventually I think both scripts should be in Python. The reason `download-hadoop` was written in Python was because of the Apache mirror selection logic, which seemed like a bit much to do purely in Bash.

flintrock/config.yaml.template
nchammas: Perhaps something like this?

```
# optional; defaults to download from official Spark S3 bucket
#   - must contain a {v} template corresponding to the version
#   - Spark must be pre-built
#   - must be a .tar.gz file
# download-source: "https://www.example.com/files/spark/{v}/spark-{v}.tar.gz"
```

And then we should update the matching comment for the Hadoop download source to follow similar formatting.

flintrock/flintrock.py
```diff
@@ -194,6 +194,10 @@ def cli(cli_context, config, provider):
               help="Git repository to clone Spark from.",
               default='https://github.com/apache/spark',
               show_default=True)
+@click.option('--spark-download-source',
```
nchammas: I suggest moving this option to right after `--spark-version` and updating the help text to read: "URL to download release version of Spark from."

That should make it slightly clearer to the user that these two options are closely related, while the two `--spark-git-...` options are separate.

flintrock/scripts/install-spark.sh
```diff
@@ -2,22 +2,20 @@
...
-distribution="$2"
+url="$1"
 
 echo "Installing Spark..."
-echo "  version: ${spark_version}"
-echo "  distribution: ${distribution}"
+echo "  from: ${url}"
 
-file="spark-${spark_version}-bin-${distribution}.tgz"
+file=$(basename ${url})
```
nchammas: Nit: We should always quote values in Bash.

I'm paranoid after having been bit too many times. 🐍

nchammas:
Thank you for this PR @BenFradet. Looks good to me! I left some minor comments.

flintrock/services.py
BenFradet: Yup, you can choose your distribution if you specify your own download source.

However, we might want to support the use case of someone only specifying the spark version and distribution. What do you think?

flintrock/config.yaml.template
BenFradet: Yup, this seems clearer

BenFradet:
Great, thanks for your review, will update accordingly.

flintrock/services.py
nchammas: Hmm, for now let's leave it like this. I have some vague concerns about "officially" supporting other distributions, in case they have annoying problems that we would have to work around. With the download source option, people who really want a different distribution can get it, and we have a bit more of an excuse to deflect support if there are serious issues.

It's definitely something I am open to revisiting in the future, though.

ereed-tesla:
This is slick -- S3 support via the Spark hadoop-2.4 is pretty convenient. Is there anything remaining to get this merged in?

I confirmed this PR works by merging into master and doing the following:
```yaml
download-source: "http://mirror.cc.columbia.edu/pub/software/apache/spark/spark-1.6.2/spark-1.6.2-bin-hadoop2.4.tgz"
```
--------
```
$ flintrock launch erik-flintrock --ec2-instance-type m3.medium
Launching 2 instances...
[52.40.196.100] SSH online.
[52.40.196.100] Configuring ephemeral storage...
[52.40.196.100] Installing HDFS...
[52.39.213.68] SSH online.
[52.39.213.68] Configuring ephemeral storage...
[52.39.213.68] Installing HDFS...
[52.39.213.68] Installing Spark...
[52.40.196.100] Installing Spark...
[52.39.213.68] Configuring HDFS master...
[52.39.213.68] Configuring Spark master...
HDFS online.
Spark Health Report:
  * Master: ALIVE
  * Workers: 1
  * Cores: 1
  * Memory: 2.7 GB            
launch finished in 0:04:00.
```
Then successfully fetching some data from S3:
```
Welcome to
      ____              __
     / __/__  ___ _____/ /__
    _\ \/ _ \/ _ `/ __/  '_/
   /__ / .__/\_,_/_/ /_/\_\   version 1.6.2
      /_/

Using Python version 3.5.1 (default, Dec  7 2015 11:16:01)
SparkContext available as sc, HiveContext available as sqlContext.   

In [1]: df = sqlContext.parquetFile('s3n://XYZ.gz.parquet')

In [2]: df.count()
```

nchammas:
Took a second look at this. Looks good to me.

And thanks @ereed-tesla for testing it out. It speeds things up for me since I can skip on testing it myself if I am already comfortable with the PR.

Merging this in.

PR #127, opened 2016-07-10
HADOOP_LIBEXEC_DIR env var

BenFradet:
This PR adds the `HADOOP_LIBEXEC_DIR` env variable necessary to use the different hadoop cli utilities like `hdfs`

I created a test cluster and tried a few hdfs commands.


nchammas:
LGTM. Thank you for following-up on this.

PR #146, opened 2016-09-08
Ensure Java 8 is installed

### LINKED ISSUE #143: Ensure Java >= 8 on launched nodes
As raised in #142, Spark 2.0 deprecates support for Java 7. Furthermore, lots of Catalyst/Tungsten optimizations are enabled by running on Java 8.

Flintrock should, by default, ensure that all launched nodes have Java 8 or higher installed and configured to be the default.

If the user uses an AMI that already has this setup, Flintrock should detect that and do nothing.

serialx:
Since Spark 2.0 Java 1.7 is deprecated. And many of the optimisations require Java 1.8. This patch ensures Java 1.8 in CentOS-like distros including Amazon Linux. It also removes any previous versions of Java to resolve any possible version conflicts.

This closes #143

I've tested this with our production deployment.

### DIFF
--- a/flintrock/core.py
+++ b/flintrock/core.py
@@ -13,6 +13,7 @@ import paramiko
 
 # Flintrock modules
 from .ssh import get_ssh_client, ssh_check_output, ssh, SSHKeyPair
+from .exceptions import SSHError
 
 FROZEN = getattr(sys, 'frozen', False)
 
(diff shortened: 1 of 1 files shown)


flintrock/core.py
```diff
@@ -482,6 +483,58 @@ def _run_asynchronously(*, partial_func: functools.partial, hosts: list):
...
 
+def get_java_major_version(client: paramiko.client.SSHClient):
+    possible_cmds = [
+        "$JAVA_HOME/bin/java -version",
+        "java -version"
+    ]
+
+    for command in possible_cmds:
+        try:
+            output = ssh_check_output(
```
nchammas: Hmm, I think this will always return an empty byte string because `java -version` spits its output on `stderr`, whereas `ssh_check_output()` only returns `stdout`.

Ultimately, I want to change `ssh_check_output()` so that it cleanly returns both `stdout` and `stderr`, but in the meantime you can fix this by updating your `possible_cmds` to redirect `stderr` to `stdout`.

```sh
$ORIGINAL_COMMAND 2>&1
```

serialx: Hmm, you're right. There should be no output!

But I've tested this code many times to confirm the output. Let me test this again.

nchammas: It will still work for the current default AMIs because no output means Java 8 gets installed, but it will do unnecessary work if Java 8 is already installed.

serialx: Actually, my main test case was Amazon Linux where Java 1.7 is installed by default. I've confirmed the function returning (1, 7).

But just in case, I'll add `2>&1` to the code.

serialx:
Updated to redirect stderr to stdout. Should work even if the behaviour of `ssh_check_output()` changes.

nchammas:
Hey @serialx, sorry about the delay in finishing up the review on this. I'll hopefully get to this sometime next week.

flintrock/core.py
nchammas: You're right. No idea why that's happening... Thanks for patching that up anyway.

flintrock/core.py
```diff
@@ -482,6 +483,58 @@ def _run_asynchronously(*, partial_func: functools.partial, hosts: list):
...
+            pass
+
+    return None
+
+
+def ensure_java8(client: paramiko.client.SSHClient):
+    host = client.get_transport().getpeername()[0]
+    java_major_version = get_java_major_version(client)
+
+    # The default CentOS AMIs on EC2 don't come with Java installed.
```
nchammas: I think we can remove this comment and the one immediately below it.

flintrock/core.py
```diff
@@ -482,6 +483,58 @@ def _run_asynchronously(*, partial_func: functools.partial, hosts: list):
...
+        ssh_check_output(
+            client=client,
+            command="""
+                set -e
+
+                # Install Java 1.8 first to protect packages that depend on Java from being removed.
+                sudo yum install -y java-1.8.0-openjdk
+
+                # Remove any older versions of Java to force the default Java to 1.8.
+                # We don't use alternatives because it does not seem to update links in /usr/lib/jvm correctly,
```
nchammas: Can we update this to say "We don't use /etc/alternatives..." or something like that so it's clear that we're talking about a specific piece software? Otherwise I will probably forget what this is about in a few months. 😆

serialx:
Done! :)

nchammas:
Excellent. Thank you for your patience!

PR #149, opened 2016-09-21
Add --ec2-user-data option to launch command

SoyeonBaek-dev:
This PR adds `--ec2-user-data` option to launch command.
This flag can help users create `UserData` defined instances. 
`--ec2-user-data` requires the path of `UserData` script file, and it could support long and complex script under 16KB.[1]

In case of increasing the number of slave instances, new slave instances are going to have the same `userdata` from the master instance.

I tested this PR by deploying cluster to AWS.

This PR is related to [#142](https://github.com/nchammas/flintrock/pull/142).
[1]: [Configuring Instances with User Data](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html)

### DIFF
--- a/flintrock/ec2.py
+++ b/flintrock/ec2.py
@@ -668,6 +668,7 @@ def _create_instances(
 
     try:
         if spot_price:
+            user_data = base64.b64encode(user_data.encode('utf-8')).decode()
             print("Requesting {c} spot instances at a max price of ${p}...".format(
                 c=num_instances, p=spot_price))
             client = ec2.meta.client
--- a/flintrock/ec2.py
+++ b/flintrock/ec2.py
@@ -3,6 +3,8 @@ import string
 import sys
 import time
 import urllib.request
+import base64
+import os
 from collections import namedtuple
 from datetime import datetime
 
@@ -247,6 +249,15 @@ class EC2Cluster(FlintrockCluster):
         )
         instance_initiated_shutdown_behavior = response['InstanceInitiatedShutdownBehavior']['Value']
 
+        response = client.describe_instance_attribute(
+            InstanceId=self.master_instance.id,
+            Attribute='userData'
+        )
+        if not response['UserData']:
+            user_data = ''
+        else:
+            user_data = response['UserData']['Value']
+
         if not self.master_instance.iam_instance_profile:
             instance_profile_arn = ''
         else:
@@ -268,7 +279,8 @@ class EC2Cluster(FlintrockCluster):
             subnet_id=self.master_instance.subnet_id,
             instance_profile_arn=instance_profile_arn,
             ebs_optimized=self.master_instance.ebs_optimized,
-            instance_initiated_shutdown_behavior=instance_initiated_shutdown_behavior)
+            instance_initiated_shutdown_behavior=instance_initiated_shutdown_behavior,
+            user_data=user_data)
 
         (ec2.instances
             .filter(
@@ -647,7 +659,8 @@ def _create_instances(
         subnet_id,
         instance_profile_arn,
         ebs_optimized,
-        instance_initiated_shutdown_behavior) -> 'List[boto3.resources.factory.ec2.Instance]':
+        instance_initiated_shutdown_behavior,
+        user_data) -> 'List[boto3.resources.factory.ec2.Instance]':
     ec2 = boto3.resource(service_name='ec2', region_name=region)
 
     cluster_instances = []
@@ -673,7 +686,8 @@ def _create_instances(
                     'SubnetId': subnet_id,
                     'IamInstanceProfile': {
                         'Arn': instance_profile_arn},
-                    'EbsOptimized': ebs_optimized})['SpotInstanceRequests']
+                    'EbsOptimized': ebs_optimized,
+                    'UserData': user_data})['SpotInstanceRequests']
 
             request_ids = [r['SpotInstanceRequestId'] for r in spot_requests]
             pending_request_ids = request_ids
@@ -728,7 +742,8 @@ def _create_instances(
                 IamInstanceProfile={
                     'Arn': instance_profile_arn},
                 EbsOptimized=ebs_optimized,
-                InstanceInitiatedShutdownBehavior=instance_initiated_shutdown_behavior)
+                InstanceInitiatedShutdownBehavior=instance_initiated_shutdown_behavior,
+                UserData=user_data)
 
         time.sleep(10)  # AWS metadata eventual consistency tax.
         return cluster_instances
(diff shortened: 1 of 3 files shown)


SoyeonBaek-dev:
I found the error when requesting the spot instances and fixed it. 
I'm testing now, and let you know when I'm done.

flintrock/flintrock.py
```diff
@@ -226,6 +226,8 @@ def cli(cli_context, config, provider):
 @click.option('--ec2-ebs-optimized/--no-ec2-ebs-optimized', default=False)
 @click.option('--ec2-instance-initiated-shutdown-behavior', default='stop',
               type=click.Choice(['stop', 'terminate']))
+@click.option('--ec2-user-data', default='',
```
nchammas: Would it be appropriate to use Click's `Path` type here? I think it would help us eliminate the need for `_get_user_data()`.

References:
* http://click.pocoo.org/6/api/#click.Path
* http://click.pocoo.org/6/arguments/#file-path-arguments

Example from Flintrock:
* https://github.com/nchammas/flintrock/blob/7e811104184e1d45e02bf0f4056fbc2d53ba2746/flintrock/flintrock.py#L206-L208

flintrock/flintrock.py
```diff
@@ -226,6 +226,8 @@ def cli(cli_context, config, provider):
...
+              help="Path to userdata script file for accessing nodes.")
```
nchammas: I don't understand what you mean here by "for accessing nodes". Perhaps we could just phrase this as follows?

> Path to EC2 user data script that will run on instance launch.

nchammas:
Ah, I just saw your most recent comment, @SoyeonBaek-dev. OK cool.

flintrock/flintrock.py
SoyeonBaek-dev: Oh, I didn't fix the typo. Thanks for checking it.

flintrock/flintrock.py
SoyeonBaek-dev: That's great. I missed that point. 
I would change it to Click's `Path` type and remove `_get_user_data()`.

flintrock/ec2.py
```diff
@@ -3,6 +3,8 @@
 import sys
 import time
 import urllib.request
+import base64
```
SoyeonBaek-dev: Yes.  As you said I didn't use `base64` module in this commit, but I need it in next commit. 

`ec2.create_instances()` supports automatically base64 encoded `UserData`, however, 
`ec2.meta.client.request_spot_instances()` doesn't seem to support that.

 I think I should give base64 encoded string `user_data` to `ec2.meta.client.request_spot_instances()` `UserData` parameter. 

This is what I'm dealing with in above 'error when requesting the spot instances' problem.

flintrock/flintrock.py
SoyeonBaek-dev: I changed `--ec2-user-data` option type to Click's `File` type, and remove `_get_user_data()`.  
I think Click's `File` type is more appropriate than Click's `Path` type since user-data script should be opened and read. 
Could you give me a comment for this?

SoyeonBaek-dev:
I fixed the bug about requesting spot instances with user-data.
Also, I changed the launch `--ec2-user-data` option type and help message. 

Please review, and feel free to give me any comments.

flintrock/flintrock.py
nchammas: `File` sounds good to me!

nchammas:
Updated PR looks good to me. Going to test this out now.

nchammas:
I'm seeing this when I try to add slaves to a regular, on-demand cluster:

```
$ flintrock add-slaves nick --num-slaves 1
  File ".../venv/bin/flintrock", line 9, in <module>
    load_entry_point('Flintrock==0.7.0.dev0', 'console_scripts', 'flintrock')()
  File ".../venv/lib/python3.5/site-packages/flintrock/flintrock.py", line 1039, in main
    cli(obj={})
  File ".../venv/lib/python3.5/site-packages/click/core.py", line 716, in __call__
    return self.main(*args, **kwargs)
  File ".../venv/lib/python3.5/site-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File ".../venv/lib/python3.5/site-packages/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File ".../venv/lib/python3.5/site-packages/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File ".../venv/lib/python3.5/site-packages/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File ".../venv/lib/python3.5/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File ".../venv/lib/python3.5/site-packages/flintrock/flintrock.py", line 661, in add_slaves
    **provider_options)
  File ".../venv/lib/python3.5/site-packages/flintrock/ec2.py", line 48, in wrapper
    res = func(*args, **kwargs)
  File ".../venv/lib/python3.5/site-packages/flintrock/ec2.py", line 283, in add_slaves
    user_data=user_data)
  File ".../venv/lib/python3.5/site-packages/flintrock/ec2.py", line 668, in _create_instances
    user_data = user_data.read()
AttributeError: 'str' object has no attribute 'read'
```

I think the issue is that `_create_instances()` should probably be passed the file contents rather than the file handle during `launch`, which would make things consistent with `add-slaves`.

flintrock/ec2.py
```diff
@@ -647,14 +659,17 @@ def _create_instances(
...
         instance_profile_arn,
         ebs_optimized,
-        instance_initiated_shutdown_behavior) -> 'List[boto3.resources.factory.ec2.Instance]':
+        instance_initiated_shutdown_behavior,
+        user_data) -> 'List[boto3.resources.factory.ec2.Instance]':
     ec2 = boto3.resource(service_name='ec2', region_name=region)
 
     cluster_instances = []
     spot_requests = []
+    user_data = user_data.read()
```
nchammas: To follow-on to my comment on the main PR, we should remove this line and have the methods calling `_create_instances()` take care of reading the file contents where appropriate.

nchammas:
@SoyeonBaek-dev - When you fix this issue, just confirm to me that you've successfully tested the following scenarios and I will merge this in:

1. Launching a regular cluster.
2. Adding slaves to a regular cluster.
3. Launching a spot cluster.
4. Adding slaves to a spot cluster.

flintrock/ec2.py
SoyeonBaek-dev: I removed that 668 line. 
Instead of that, reading the user-data file contents is located in the `launch` before calling `_create_instances()` if `--ec2-user-data` is given.

SoyeonBaek-dev:
I should have tested all the test in the flintrock, and now I passed all the test except `test_launch_with_bad_ami` test(this test should be fixed).
Also, I checked all the scenarios you mentioned above. 

Please check and test once again.

nchammas:
Retested this and it all works as expected now. Thank you @SoyeonBaek-dev!

PR #152, opened 2016-10-02
Add Qubole to list of managed Spark providers

jsensarma:
* Adds Qubole link to README


nchammas:
Hi @jsensarma and thank you for this PR.

I'd like to keep the text as-is and not add any more examples. The README is not meant to be an exhaustive list of Spark-related services, but rather it is meant to be a sampling of alternatives the reader may be interested in.

I mentioned EMR since Flintrock currently only supports EC2, and EMR is a natural comparison to make to Filntrock within the EC2 ecosystem. And I mentioned Databricks since they are the main driving force behind Spark itself and were probably the first to offer a hosted Spark service.

Qubole is probably a great product too (I've heard good things about them), but at this point I'm not interested in modifying the README to add them or any other Spark-as-a-service products.

PR #153, opened 2016-10-10
Fix handling of AMI not found

nchammas:
This fixes the `test_launch_with_bad_ami` test (and the user-facing error message).

It appears that at some point this year Amazon changed the behavior of their API to never throw `InvalidAMIID.NotFound` when using `ec2.images.Filter`.


nchammas:
Thank you for bringing this issue to my attention @SoyeonBaek-dev.

PR #162, opened 2016-11-04
Update OpenSSL before starting Travis CI build

nchammas:


nchammas:
Misguided attempt at addressing #158.

PR #164, opened 2016-11-07
Resolve #163. Allow more ports for SparkUI.

### LINKED ISSUE #163: Allow Ports Ranging 4040 ~ 4050
Current flintrock [only allows public access on port 4040][1].  But when there is more than one Spark applications running, additional ports are allocated for SparkUI.  [Default port are 4040][2] in the SparkUI code.  But, [when the default port is already allocated, it increments the port until bind is successful][3].

I think we should allow port range 4040 ~ 4050.

I'll prep a PR soon. Does this makes sense?

[1]: https://github.com/nchammas/flintrock/blob/master/flintrock/ec2.py#L544
[2]: https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/ui/SparkUI.scala#L137
[3]: https://github.com/apache/spark/blob/70a5db7bbd192a4bc68bcfdc475ab221adf2fcdd/core/src/main/scala/org/apache/spark/ui/JettyUtils.scala#L266

serialx:
Resolves #163.  When there is more than one Spark applications running, additional ports
are allocated for SparkUI.  [Default port are 4040][1] in the SparkUI
code.  But, [when the default port is already allocated, it increments
the port until bind is successful][2].

[1]: https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/ui/SparkUI.scala#L137
[2]: https://github.com/apache/spark/blob/70a5db7bbd192a4bc68bcfdc475ab221adf2fcdd/core/src/main/scala/org/apache/spark/ui/JettyUtils.scala#L266

### DIFF
--- a/flintrock/ec2.py
+++ b/flintrock/ec2.py
@@ -544,7 +544,7 @@ def get_or_create_flintrock_security_groups(
         SecurityGroupRule(
             ip_protocol='tcp',
             from_port=4040,
-            to_port=4040,
+            to_port=4050,
             cidr_ip=flintrock_client_cidr,
             src_group=None),
         SecurityGroupRule(


nchammas:
I think to link this PR back to the issue such that the issue gets close automatically, you need to have something like "Fixes #163" or "Resolves #163" in the PR description, not the title.

As for the broken tests, it looks like `pycodestyle` made a new release and started checking for new things. Guess I should [pin that dependency](https://github.com/nchammas/flintrock/blob/65160b4b8d1a603d7e9f7a9ca150be82f09fd792/requirements/developer.pip#L2) too to avoid sudden breakages like this...

serialx:
Changed the PR description. I thought that the issue was closed via commit messages, not PR descriptions. I'm not completely sure though.

See: https://help.github.com/articles/closing-issues-via-commit-messages/

nchammas:
Oh, hmm. I guess you're right, the commit message should work too.

PR #176, opened 2017-01-20
Make it work for root or users with non-standard home directories

douglaz:
- Enable root user or users with non-standard home directories to be used as --ec2-user

I tested this PR by manually launching a cluster.

Note: This won't fix this TODO:
```bash
# TODO: Make this dependent on HDFS install.
export HADOOP_CONF_DIR="/home/$(logname)/hadoop/conf"
```

### DIFF
--- a/flintrock/core.py
+++ b/flintrock/core.py
@@ -140,7 +140,7 @@ class FlintrockCluster:
             manifest_raw = ssh_check_output(
                 client=master_ssh_client,
                 command="""
-                    cat /home/{u}/.flintrock-manifest.json
+                    cat ~{u}/.flintrock-manifest.json
                 """.format(u=shlex.quote(user)))
             # TODO: Would it be better if storage (ephemeral and otherwise) was
             #       implemented as a Flintrock service and tracked in the manifest?
@@ -609,8 +609,8 @@ def provision_cluster(
         ssh_check_output(
             client=master_ssh_client,
             command="""
-                echo {m} > /home/{u}/.flintrock-manifest.json
-                chmod go-rw /home/{u}/.flintrock-manifest.json
+                echo {m} > ~{u}/.flintrock-manifest.json
+                chmod go-rw ~{u}/.flintrock-manifest.json
             """.format(
                 m=shlex.quote(json.dumps(manifest, indent=4, sort_keys=True)),
                 u=shlex.quote(user)))


flintrock/core.py
```diff
@@ -140,8 +140,8 @@ def load_manifest(self, *, user: str, identity_file: str):
             manifest_raw = ssh_check_output(
                 client=master_ssh_client,
                 command="""
-                    cat /home/{u}/.flintrock-manifest.json
-                """.format(u=shlex.quote(user)))
```
nchammas: Why are we removing the `shlex.quote()`? It's there to protect against funny business with the user name.

douglaz: You're right. I thought it would always quote the username which would make this change ineffective. My mistake. I will fix it.

douglaz:
@nchammas, I'm testing this script as a replacement to spark-ec2. As spark-ec2 installs stuff on /root by default and there is some software in our stack that takes this as a convention, I'm trying to keep things smooth for a potential transition.

douglaz:
@nchammas, yes, I tested and it ran a simple spark job that uses HDFS. But I won't guarantee that this installation is 100% functional because I'm couldn't run yet our most complete and demanding jobs due to some differences between this and spark-ec2. I'm just tackling one problem at the time and this was one of the first.

nchammas:
Okie doke, sounds good to me! I'll test this over the weekend and merge in.

nchammas:
@douglaz - Some quick questions for you.

1. I see that we still have two instances of `/home/$(logname)` in `spark-env.sh` and `hadoop-env.sh`. Shouldn't those also be updated to `~`? I assume otherwise this would break on users who have a non-standard home, no?
2. If I do `echo ~` I see that `~` expands to `/home/ec2-user` for the `ec2-user`. That means the `{u}` in the current diff is unnecessary and can be dropped, right?

douglaz:
1. I agree that it's wrong, but changing it to ~ won't fix it. It's because you don't know who is going to load these environment variables. In our system, for instance, everything is installed under /root but there is a standard user (ec2-user) who's going to read these variables. If ~ is used, this user will think that stuff is installed under his home, which is false. You need to use something like ~{u}, which is an absolute value. Or, better, just define a single time where the spark/hdfs will be installed and use the the full expanded path wherever it's needed
2. This works if you guarantee the same system user is the one running both commands. This is a reasonable assumption and, AFAIK, true according to current code. But using ~ can be dangerous as shown above because you need to taking in consideration all users that may evaluate this code and guarantee that it's always the same user. I think it's easier and fool-proof to user absolute paths like ~{u} or /tmp (in our system we do something very similar to save metadata about the cluster and we use /tmp)

nchammas:
1. So to address this problem I believe we need to make the installing user part of the template for `spark-env.sh` and `hadoop-env.sh` (like, for example, `{master_host}`). Does that make sense?
2. You are correct that we assume the user will stick to one username throughout the cluster's lifecycle, but since as you explained `~{u}` is more robust let's leave it that way. 👍

douglaz:
1. Yes, that would be good

nchammas:
OK, I can merge this in and make that change if you want @douglaz.

nchammas:
Or, alternately, I can [directly push to your branch](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) since you've allowed that in this PR.

douglaz:
@nchammas, feel free to push to my branch

nchammas:
Sorry about the delay here. I'm working on a PR that touches many of the templates. Once I get that in I'll revisit this and get it merged in.

nchammas:
Now that #180 is in, I'll get this in next.

PR #185, opened 2017-02-01
Added /media/tmp for temporary usage that needs more space than /tmp

douglaz:
This PR makes creates a /media/tmp linking to /ephemeral*

A usage example is to set SPARK_WORKER_DIR to this.

I tested this PR by launching a cluster.

### DIFF
--- a/flintrock/scripts/setup-ephemeral-storage.py
+++ b/flintrock/scripts/setup-ephemeral-storage.py
@@ -12,7 +12,7 @@ The resulting structure we create is as follows:
         /root: The instance's root volume.
         /ephemeral[0-N]: Instance store volumes.
         /persistent[0-N]: EBS volumes.
-        /tmp: A temporary directory with lots of space (either /ephemeral* or /persistent*)
+        /tmp: A temporary directory with lots of space.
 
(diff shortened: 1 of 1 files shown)


flintrock/scripts/setup-ephemeral-storage.py
```diff
@@ -159,6 +160,23 @@ def create_root_dir():
     return path
 
 
+def create_tmp_dir(devices):
```
nchammas: Selecting `devices[0]` a few lines down smells a bit to me. I think this API would be clearer if it simply accepted the path that `/media/tmp` should be pointed at. `__main__` should be tasked with selecting the appropriate path to pass to this function.

douglaz:
> Some testing questions:
> 
> Did you test this with instances that don't have instance store volumes (e.g. t2.medium)?
Yes, tested on m4.large
> Did you test this with instances that do have instance store volumes (e.g. m3.xlarge)?
Yes, this was my first test
> Did you test this with adding nodes to an existing cluster?
Yep

nchammas:
Sorry about the delay in merging this in @douglaz and thank you for the contribution.

PR #186, opened 2017-02-07
add option to tag ec2 instances

### LINKED ISSUE #182: Adding custom tags
First off, thank you for this awesome package. I've been using it extensively for the last couple of months and it's been wonderful.

At my work, we label launched ec2 instances with additional tags for book-keeping purposes. Is there any interest to add that functionality to flintrock? Right now, the only tags flintrock uses are flintrock-role and Name and there's no way to append custom tags. I'm happy to take a stab if you think it's a worthwhile addition.

lwoloszy:
This PR allows users to label their EC2 instances with custom tags. An EC2 tag is added with the --ec2-tag option, where each tag is a Key,Value pair separated by a comma. The ec2-tag can be specified multiple times.

I tested this both for launching new clusters and appending slaves to existing clusters. As far as I can tell, it works as expected.

Speaking of the code itself, there's a common formatting step that takes the tags as supplied on the command line and transforms them into a dictionary that EC2 expects. This piece of code is replicated for launching and adding slaves as I didn't know the proper place to put it if we wanted to wrap it as a function (the ec2.py file?). This would be trivial to do if desired.

Cheers,
Luke

Fixes #182

### DIFF
--- a/flintrock/config.yaml.template
+++ b/flintrock/config.yaml.template
@@ -37,8 +37,8 @@ providers:
     #   - group-name2
     # instance-profile-name:
     # tags:
-    #   - key1,value1
-    #   - key2,value2
+    #   - key1,value1 (or 'key1, value1' or 'key1,')
+    #   - key2,value2 (or 'key2, value2' or 'key2,')
     tenancy: default  # default | dedicated
     ebs-optimized: no  # yes | no
     instance-initiated-shutdown-behavior: terminate  # terminate | stop
--- a/flintrock/config.yaml.template
+++ b/flintrock/config.yaml.template
@@ -36,6 +36,9 @@ providers:
     #   - group-name1
     #   - group-name2
     # instance-profile-name:
+    # tags:
+    #   - key1,value1
+    #   - key2,value2
     tenancy: default  # default | dedicated
     ebs-optimized: no  # yes | no
     instance-initiated-shutdown-behavior: terminate  # terminate | stop
--- a/flintrock/config.yaml.template
+++ b/flintrock/config.yaml.template
@@ -37,8 +37,9 @@ providers:
     #   - group-name2
     # instance-profile-name:
     # tags:
-    #   - key1,value1 (or 'key1, value1' or 'key1,')
-    #   - key2,value2 (or 'key2, value2' or 'key2,')
+    #   - key1,value1
+    #   - key2, value2  # leading/trailing spaces are trimmed
+    #   - key3,  # value will be empty
     tenancy: default  # default | dedicated
     ebs-optimized: no  # yes | no
     instance-initiated-shutdown-behavior: terminate  # terminate | stop
--- a/flintrock/ec2.py
+++ b/flintrock/ec2.py
@@ -234,6 +234,7 @@ class EC2Cluster(FlintrockCluster):
             identity_file: str,
             num_slaves: int,
             spot_price: float,
+            tags: list,
             assume_yes: bool):
         security_group_ids = [
             group['GroupId']
@@ -287,15 +288,17 @@ class EC2Cluster(FlintrockCluster):
                 instance_initiated_shutdown_behavior=instance_initiated_shutdown_behavior,
                 user_data=user_data)
 
+            slave_tags = [
+                {'Key': 'flintrock-role', 'Value': 'slave'},
+                {'Key': 'Name', 'Value': '{c}-slave'.format(c=self.name)}]
+            slave_tags += tags
+
             (ec2.instances
                 .filter(
                     Filters=[
                         {'Name': 'instance-id', 'Values': [i.id for i in new_slave_instances]}
                     ])
-                .create_tags(
-                    Tags=[
-                        {'Key': 'flintrock-role', 'Value': 'slave'},
-                        {'Key': 'Name', 'Value': '{c}-slave'.format(c=self.name)}]))
+                .create_tags(Tags=slave_tags))
 
             existing_slaves = {i.public_ip_address for i in self.slave_instances}
 
@@ -817,7 +820,8 @@ def launch(
         tenancy='default',
         ebs_optimized=False,
         instance_initiated_shutdown_behavior='stop',
-        user_data):
+        user_data,
+        tags):
     """
     Launch a cluster.
     """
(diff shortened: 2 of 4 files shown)


flintrock/config.yaml.template
```diff
@@ -36,6 +36,9 @@ providers:
     #   - group-name1
     #   - group-name2
     # instance-profile-name:
+    # tags:
+    #   - key1,value1
```
nchammas: Would `key1, value1` work? (i.e. with a space)

flintrock/flintrock.py
```diff
@@ -236,6 +236,10 @@ def cli(cli_context, config, provider):
 @click.option('--ec2-user-data',
               type=click.File(mode='r', encoding='utf-8'),
               help="Path to EC2 user data script that will run on instance launch.")
+@click.option('--ec2-tag', 'ec2_tags',
+              multiple=True,
+              help="Additional tags (Key,Value pairs) to assign to the instances. "
```
nchammas: Nitpick: I think this would be slightly clearer if it were written as: "Additional tags (e.g. 'Key,Value') to assign to the instances. "

flintrock/flintrock.py
```diff
@@ -334,6 +339,22 @@ def launch(
             )
         services += [spark]
 
+    # Format ec2 tags
```
nchammas: This should be split out into a function instead of being duplicated here and in `add_slaves()`. Then, that function should be used to validate the input right as it's parsed by Click:

http://click.pocoo.org/6/options/#callbacks-for-validation

flintrock/flintrock.py
```diff
@@ -334,6 +339,22 @@ def launch(
...
+    if ec2_tags:
+        for ec2_tag in ec2_tags:
+            if ec2_tag.count(',') != 1:
+                raise UsageError(
+                    "Error: EC2 tags need to be specified as Key,Value pairs "
+                    "separated by a single comma")
+            k, v = ec2_tag.split(',')
+            if not len(k) or not len(v):
+                raise UsageError(
+                    "Error: An EC2 tag seems to have a missing key or value")
```
nchammas: Is a missing value a problem? I would assume that you can tag something with a key but no value.

flintrock/flintrock.py
```diff
@@ -334,6 +339,22 @@ def launch(
             )
         services += [spark]
 
+    # Format ec2 tags
+    if ec2_tags:
+        for ec2_tag in ec2_tags:
+            if ec2_tag.count(',') != 1:
+                raise UsageError(
```
nchammas: If you convert this into a function that gets called via a Click callback, it might be good to replace this with [`BadParameter`](http://click.pocoo.org/6/api/#click.BadParameter).

flintrock/flintrock.py
```diff
@@ -334,6 +339,22 @@ def launch(
...
+                    "Error: EC2 tags need to be specified as Key,Value pairs "
+                    "separated by a single comma")
```
nchammas: Nit: Please end prose with proper punctuation.

flintrock/flintrock.py
nchammas: Once it's split out into a function, we should probably also add a test or two to `tests/test_ec2.py`.

lwoloszy:
Hi @nchammas ,

Glad to contribute! And thanks for the feedback. I've added the changes as requested (which have def streamlined the code). One thing I'm still unsure of is how to allow for tags that have only a Key, but no Value. As it's implemented right now, if a user wants a tag with only a Key, the comma separating the Key from the (implicitly) empty Value is still required. We may want to allow the option to specify a Key without the comma. Once we've made that decision, I'm happy to add a couple of tests.

Luke

flintrock/config.yaml.template
```diff
@@ -37,8 +37,8 @@ providers:
     #   - group-name2
     # instance-profile-name:
     # tags:
-    #   - key1,value1
-    #   - key2,value2
+    #   - key1,value1 (or 'key1, value1' or 'key1,')
```
nchammas: I would put the examples on separate lines with a clarifying comment:

```
tags:
  - key1,value1
  - key2, value2  # leading/trailing spaces are trimmed
  - key3,  # value will be empty
```

flintrock/ec2.py
```diff
@@ -1014,6 +1014,24 @@ def get_clusters(*, cluster_names: list=[], region: str, vpc_id: str) -> list:
     return clusters
 
 
+def validate_tags(ctx, param, value):
+    """
+    Validate and parse optional ec2 tags as supplied on command line.
```
nchammas: I would just phrase this as "Validate and parse optional EC2 tags." since they can be specified via the config file too.

flintrock/ec2.py
```diff
@@ -1014,6 +1014,24 @@ def get_clusters(*, cluster_names: list=[], region: str, vpc_id: str) -> list:
...
+    """
+    ec2_tags = value
+    for ec2_tag in ec2_tags:
+        if ec2_tag.count(',') != 1 or not ec2_tag.split(',')[0]:
+            raise click.BadParameter(
+                "tags need to be specified as 'Key,Value' pairs "
```
nchammas: Nit: "Tags need to be..." (capitalization)

flintrock/ec2.py
```diff
@@ -1014,6 +1014,24 @@ def get_clusters(*, cluster_names: list=[], region: str, vpc_id: str) -> list:
...
+                "separated by a single comma. Key cannot be empty.")
+
+    ec2_tags = [{'Key': tag.split(',')[0].strip(),
+                 'Value': tag.split(',')[1].strip()}
+                for tag in ec2_tags]
```
nchammas: We won't catch keys that are whitespace: `--ec2-tag '   ,ihavenokeylulz'`

Is there a way to unify the logic so that we parse the tags only one time, instead of once for error checking and then again to create the data structure we need? That way we avoid these issues.

Perhaps something like

```python
key, value = [word.strip() for word in tag.split(',', maxsplit=1)]
```

followed by a) whatever error checking we need to do, and then b) using `key` and `value` to build the data structure we need. That way we parse only once.

flintrock/flintrock.py
```diff
@@ -237,9 +237,9 @@ def cli(cli_context, config, provider):
               type=click.File(mode='r', encoding='utf-8'),
               help="Path to EC2 user data script that will run on instance launch.")
 @click.option('--ec2-tag', 'ec2_tags',
+              callback=ec2.validate_tags,
               multiple=True,
-              help="Additional tags (Key,Value pairs) to assign to the instances. "
-                   "You can specify this option multiple times.")
```
nchammas: Why did you remove this second sentence? I think it's helpful.

flintrock/flintrock.py
```diff
@@ -642,9 +626,9 @@ def stop(cli_context, cluster_name, ec2_region, ec2_vpc_id, assume_yes):
 @click.option('--ec2-spot-price', type=float)
 @click.option('--assume-yes/--no-assume-yes', default=False)
 @click.option('--ec2-tag', 'ec2_tags',
+              callback=ec2.validate_tags,
               multiple=True,
-              help="Additional tags (Key,Value pairs) to assign to the instances. "
-                   "You can specify this option multiple times.")
```
nchammas: Same here. I'd restore the second sentence. Seems helpful to me. Sorry if I mislead you into removing it the first time around.

lwoloszy:
Cool, I think we're almost there.  Just to make sure, I don't think we want to allow keys that are all whitespace. I've made that explicit in the error message that's thrown. 

Whoops, forgot a test case before pushing. if we try to split on a comma before checking that a comma exists, bad things ensue. Perhaps that's why I had decided to do some error checking before parsing the arguments. Let me think about it.

flintrock/ec2.py
```diff
@@ -1016,20 +1016,22 @@ def get_clusters(*, cluster_names: list=[], region: str, vpc_id: str) -> list:
...
-                for tag in ec2_tags]
-
-    return ec2_tags
+    err_msg = ("Tags need to be specified as 'Key,Value' pairs "
+               "separated by a single comma. Key cannot be empty "
+               "or be made up entirely of whitespace.")
+    tags = value
+    result = []
+    for tag in tags:
+        if ',' not in tag:
```
nchammas: How about `if tag.count(',') != 1:` over here...

flintrock/ec2.py
```diff
@@ -1016,20 +1016,22 @@ def get_clusters(*, cluster_names: list=[], region: str, vpc_id: str) -> list:
...
+            raise click.BadParameter(err_msg)
+        key, value = [word.strip() for word in tag.split(',', maxsplit=1)]
+        if not key or ',' in value:
```
nchammas: ...and simply `if not key:` over here?

lwoloszy:
Great, logic simplified.

As far as the tests go, should I put them in a separate test_ec2.py file?

nchammas:
> As far as the tests go, should I put them in a separate test_ec2.py file?

Yup.

lwoloszy:
Hi @nchammas,

Here are a couple versions of tests for the ec2 validate_tags function. Not sure which kind you prefer, so I've included both. Not an expert on writing tests, so let me know where things can be improved.

Cheers,
Luke

lwoloszy:
Happy to help out!

PR #191, opened 2017-03-13
Let the min root ebs size be configurable

### LINKED ISSUE #174: Option for Minimum EBS Root Volume Size
* Flintrock version: 0.7.0

There should be an option to modify `min_root_device_size_gb = 30` in line 626, ec2.py to any desired value in the flintrock configuration file. 30 GB may be excessive and costly in some cases, provided the AMI is smaller than 30 GB (10 GB in my case).

Edit: I address this also in my [guide](https://github.com/PiercingDan/spark-Jupyter-AWS).

douglaz:
It was hard coded. Now it's configurable. This is useful for m4, r4, c4 machines without ephemeral disks.

Fixes #174.

### DIFF
--- a/flintrock/config.yaml.template
+++ b/flintrock/config.yaml.template
@@ -40,6 +40,7 @@ providers:
     #   - key1,value1
     #   - key2, value2  # leading/trailing spaces are trimmed
     #   - key3,  # value will be empty
+    # min-root-ebs-size-gb: <size-gb>
     tenancy: default  # default | dedicated
     ebs-optimized: no  # yes | no
     instance-initiated-shutdown-behavior: terminate  # terminate | stop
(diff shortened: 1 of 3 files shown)


douglaz:
Let's see how it works in production. Then we can change the other clusters (daily, etc)

nchammas:
I initially resisted adding an option for this in #174, but with the additional detail here explaining why it would be useful I am convinced. (I edited a reference to #174 into the PR description btw.)

Tested this out with `launch` and `add-slaves`. Looks good to me.

Only thing missing is that the new option needs to be added to the config template. Add that in and I can merge this PR. 👍

nchammas:
Never mind, I figured out how to push directly to your branch. Merging this in. Thanks!

PR #202, opened 2017-07-03
User templates for hadoop / spark

### LINKED ISSUE #200: Hadoop / Spark conf templates
I really like Flintrock as I was able to completely set up my EC2 cluster with relative ease.

One thing that would be nice is to handle user templated configuration files for hadoop and spark.
I was thinking of adding a field in config.yaml for the templates folder path.

If it sounds good to you, I can make a pull request.

ncherel:
* add a field in the the configuration file to allow user to use his own templates for hadoop / spark

I manually tested this PR. 
I started to run `pytest` but encountered errors not related to my changes. test_static are okay
I will run the full test set once the general direction of this PR is approved.

I was thinking of adding a `flintrock generate-templates /path/to/folder` command to provide the base templates (those bundled with Flintrock). What do you think?

Fixes #200

### DIFF
--- a/flintrock/services.py
+++ b/flintrock/services.py
@@ -179,7 +179,7 @@ class HDFS(FlintrockService):
                                 spark_version='',
                                 spark_executor_instances=0,
                             ))),
-                    p=shlex.quote(template_path)))
+                    p=shlex.quote(os.path.join("hadoop/conf/",template_path))))
 
     # TODO: Convert this into start_master() and split master- or slave-specific
     #       stuff out of configure() into configure_master() and configure_slave().
@@ -342,7 +342,7 @@ class Spark(FlintrockService):
                                 hadoop_version=self.hadoop_version,
                                 spark_version=self.version or self.git_commit,
                             ))),
-                    p=shlex.quote(template_path)))
+                    p=shlex.quote(os.path.join("spark/conf/", template_path))))
 
     # TODO: Convert this into start_master() and split master- or slave-specific
     #       stuff out of configure() into configure_master() and configure_slave().
--- a/flintrock/services.py
+++ b/flintrock/services.py
@@ -107,9 +107,10 @@ class FlintrockService:
 
 
 class HDFS(FlintrockService):
-    def __init__(self, *, version, download_source):
+    def __init__(self, *, version, download_source, config_path):
         self.version = version
         self.download_source = download_source
+        self.config_path = config_path
         self.manifest = {'version': version, 'download_source': download_source}
 
     def install(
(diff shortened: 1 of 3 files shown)


nchammas:
Hi @NCherel and thank you for this contribution to Flintrock! Sorry about the delay in reviewing it. I will try to take a detailed look through and give feedback by the end of this week.

> I was thinking of adding a `flintrock generate-templates /path/to/folder` command to provide the base templates (those bundled with Flintrock). What do you think?

I would prefer not to add new commands and instead enhance `flintrock configure` to handle this. Perhaps we can duplicate the template tree into Flintrock's config directory (like we currently do for `config.yaml`) and that way users can simply find the templates with `flintrock configure --locate` and edit them as they please.

I'll have to think about this more, but that's my initial reaction.

flintrock/services.py
```diff
@@ -107,10 +107,14 @@ def health_check(
...
-    def __init__(self, *, version, download_source):
+    def __init__(self, *, version, download_source, template_dir):
         self.version = version
         self.download_source = download_source
-        self.manifest = {'version': version, 'download_source': download_source}
+        self.template_dir = template_dir
+        self.manifest = {
+            'version': version,
+            'download_source': download_source,
+            'template_dir': template_dir}
```
nchammas: This raises an important issue: The point of the manifest is to be a self-contained representation of how the cluster is configured. Any instance of Flintrock should be able to manage a cluster solely using the information in the manifest and from EC2.

With this change, however, Flintrock needs access to the specified local directory to get the correct templates. So if one instance of Flintrock launches a cluster, another instance of Flintrock on a different machine won't be able to expand or restart it correctly because the manifest only has the paths to the templates on some non-cluster machine and not the templates themselves.

Realistically, I would guess that most people using Flintrock do so from a single machine. But I'm not comfortable breaking the implicit guarantee that any instance of Flintrock at version X on any machine can fully manage clusters launched by Flintrock at version X. And I know there are some teams using Flintrock where this would matter.

To address this issue, we need to put the full contents of all the templates in the manifest (or somehow otherwise on the cluster) and use that in the `add-slaves`, `remove-slaves`, and `start` commands as necessary.

nchammas:
Hi @ncherel! Are you still interested in moving this PR forward? I think it's solid feature add to Flintrock, and if we resolve the issue with the manifest I'd be happy to merge it in. What do you think?

flintrock/services.py
nchammas: A different approach we could take here is to note the template directory and some kind of hash of the contents in the manifest. If someone wants to expand an existing Flintrock cluster without having the same template files that were used during launch, Flintrock will display a warning but try to continue anyway.

This kind of forgiving behavior is technically dangerous, but in practice it should work well. And the warning should make it clear to users that they are responsible for any borked operations since they are not using the same files that were used during launch.

We should probably use this kind of approach for the other places where Flintrock references local files during launch, like `--ec2-user-data`.

nchammas:
Closing this PR due to inactivity. I might pick this idea up later because I think it's really good!

PR #208, opened 2017-07-29
Give the user a hint when we get an SSH protocol error

### LINKED ISSUE #171: Give the user a hint about what might be wrong when they get an SSH protocol banner error
When first setting up and using Flintrock, many users hit this error:

```
Exception: Error reading SSH protocol banner[Errno 104] Connection reset by peer
```

Often the issue is related to the username, private key, or key pair being used. It would be helpful to provide this as a hint to help users solve the problem themselves.

See #65, #130, and #170.

### LINKED ISSUE #207: SSHing into an instance with the wrong username results in a SSHException
Launching a new cluster with the wrong username set by the `--ec2-user` option fails with a seemingly-unrelated `paramiko.ssh_exception.SSHException: not a valid EC private key file` error.

This because key handling in Paramiko is terribad (not my words): paramiko/paramiko#387.

We can probably safely catch this particular error and raise a more informative one because EC2 key-based auth doesn't even work with ECDSA keys.

Steps to replicate: `flintrock launch --ec2-ami=<Amazon Linux AMI ID> --ec2-user=theboss cluster-mcclusterface`
* Flintrock version: 0.8, also 0.9dev
* Python version: 3.6.1
* OS: MacOS Sierra

nchammas:
Fixes #171.
Fixes #207.

### DIFF
--- a/flintrock/ssh.py
+++ b/flintrock/ssh.py
@@ -96,6 +96,12 @@ def get_ssh_client(
         except paramiko.ssh_exception.AuthenticationException as e:
             logger.debug("[{h}] SSH AuthenticationException.".format(h=host))
             time.sleep(5)
+        except paramiko.ssh_exception.SSHException as e:
+            raise SSHError(
+                host=host,
+                message="SSH protocol error. Possible causes include using "
+                "the wrong key file or username.",
+            ) from e
     else:
         raise SSHError(
             host=host,


nchammas:
@ktdrv - If you could, please test this PR out and see if it provides the appropriate hint in the situation you described in #207.

ktdrv:
@nchammas This works in my use case -- even if it's not the clearest of error messages. That being said, I can't think of a better way to do it where efforts wouldn't be better spent on fixing the underlying issue in Paramiko.

PR #209, opened 2017-08-01
Add support for slave groups

ktdrv:
This PR makes the following change:
* Adds support for slave groups, which allow the user to tag instances created with `add-slaves` and `launch` and reference those tags later when calling `remove-slaves`.

This is an improvement over the current behavior, which follows a simple LIFO methodology, prioritizing spot instances.

I tested this PR by manually testing it over a range of input sequences. Could not run the test suite.

### DIFF
--- a/flintrock/ec2.py
+++ b/flintrock/ec2.py
@@ -7,6 +7,7 @@ import base64
 import logging
 from collections import namedtuple
 from datetime import datetime
+from itertools import groupby
 
 # External modules
 import boto3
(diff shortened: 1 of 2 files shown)


nchammas:
Hi @ktdrv! Thank you for this contribution and sorry about the delay in reviewing it.

Can you elaborate a bit on the motivation for this change? On first blush it doesn't seem worth the (minor) added complexity since the current behavior of deleting spot instances first seems like it covers typical use cases very well.

nchammas:
I'm going to close this, but if you want to continue the discussion, please do!

PR #211, opened 2017-09-13
Spark worker cleanup added.

umitunal:
* 
* 

I tested this PR by...

Fixes #NN.
Fixes #MM.


nchammas:
Hi @umitunal and thanks for making your first contribution to Flintrock!

I think instead of adding one-off configurations like this, I'd rather get the work in #202 finished up. That way, users have a well-defined way of providing their own templates to use with Flintrock.

Does that work for you?

nchammas:
Closing in favor of #202. Feel free to comment or reopen if you disagree!

PR #231, opened 2017-12-15
Wait for terminated state while destroying a cluster

### LINKED ISSUE #229: Add an option to wait for cluster destruction
At the moment flintrock doesn't wait for instances termination. It could be useful to add an option (or even force it) to wait for instances to end up at terminated state.

alexandrnikitin:
* Add `--no-wait` option to `destroy` command with False default
* `destroy` command waits for instances to be in "terminated" state  `cluster.wait_for_state('terminated') `

I tested this PR by manually launching and destroying clusters. I'm not sure about tests for this, should I add one to cover `--no-wait` option?

Fixes #229

### DIFF
--- a/flintrock/flintrock.py
+++ b/flintrock/flintrock.py
@@ -437,8 +437,9 @@ def get_latest_commit(github_repository: str):
 @click.option('--assume-yes/--no-assume-yes', default=False)
 @click.option('--ec2-region', default='us-east-1', show_default=True)
 @click.option('--ec2-vpc-id', default='', help="Leave empty for default VPC.")
+@click.option('--no-wait', default=False)
 @click.pass_context
-def destroy(cli_context, cluster_name, assume_yes, ec2_region, ec2_vpc_id):
+def destroy(cli_context, cluster_name, assume_yes, ec2_region, ec2_vpc_id, no_wait):
     """
     Destroy a cluster.
     """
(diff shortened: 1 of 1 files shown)


nchammas:
Hey @alexandrnikitin, thanks for submitting this PR! After thinking about this a bit, I think it's fine to just change the default behavior to wait for termination. I'd do that right after [here](https://github.com/nchammas/flintrock/blob/433c4412c41fe42e5ee0787b459d708c3ae39d01/flintrock/ec2.py#L181).

To explain my thought process, `wait_for_state()` is a method on the EC2 implementation of `FlintrockCluster`, so we shouldn't use it in the provider-agnostic code in `flintrock.py` or `core.py`. Instead of figuring out how to enhance the internal API to capture this new feature correctly and delineate between EC2-specific and provider-agnostic code, I think it's fine to just change the default behavior.

How long does the additional wait typically take for a 10-node cluster? If it's a long time, perhaps we should tweak the text [here](https://github.com/nchammas/flintrock/blob/433c4412c41fe42e5ee0787b459d708c3ae39d01/flintrock/flintrock.py#L467) to inform users that the operation may take some time to complete.

alexandrnikitin:
Ok, sounds good to me. I'll create another PR for that.

>How long does the additional wait typically take for a 10-node cluster?

It's around a dozen of seconds. I think the message we have right now is enough.

nchammas:
Okie doke. Just FYI for the future, you don't need to open a new PR (#232). Additional commits on the same PR are OK since they will all be squashed when I merge.

PR #232, opened 2017-12-19
Wait for terminated state while destroying a cluster

### LINKED ISSUE #229: Add an option to wait for cluster destruction

alexandrnikitin:
* destroy command waits for the instances to be in "terminated" state via `cluster.wait_for_state('terminated')`

Fixes #229


nchammas:
Thanks @alexandrnikitin!

Issue #198, opened 2017-05-05
cluster launch fail randomly


pragnesh:
* Flintrock version: current master branch
* Python version: 3.4.3
* OS: ubuntu 14.04
When tried to launch flintrock cluster it fail randomly, and if i tried again it able to launch.
Here is log,

```
ubuntu@ip-172-30-0-42:/flintrock_config$ flintrock --config config.yaml launch test                                                                           
Requesting 2 spot instances at a max price of $0.5...
0 of 2 instances granted. Waiting...
All 2 instances granted.
[54.179.153.219] SSH online.
[54.179.153.219] Configuring ephemeral storage...
[54.179.153.219] Installing Java 1.8...
[54.179.153.219] Installing HDFS...
[52.77.209.225] SSH online.
[52.77.209.225] Configuring ephemeral storage...
[52.77.209.225] Installing Java 1.8...
[54.179.153.219] Installing Spark...
[52.77.209.225] Installing HDFS...
[52.77.209.225] Installing Spark...
Do you want to terminate the 2 instances created by this operation? [Y/n]: Y
Terminating instances...
[ec2-54-179-153-219.ap-southeast-1.compute.amazonaws.com] Could not connect via SSH.
ubuntu@ip-172-30-0-42:/flintrock_config$ flintrock --config config.yaml launch test
Requesting 2 spot instances at a max price of $0.5...
0 of 2 instances granted. Waiting...
All 2 instances granted.
[52.221.209.187] SSH online.
[52.221.209.187] Configuring ephemeral storage...
[52.221.209.187] Installing Java 1.8...
[54.255.211.120] SSH online.
[54.255.211.120] Configuring ephemeral storage...
[54.255.211.120] Installing Java 1.8...
[52.221.209.187] Installing HDFS...
[54.255.211.120] Installing HDFS...
[52.221.209.187] Installing Spark...
[54.255.211.120] Installing Spark...
[172.30.0.17] Configuring HDFS master...
[172.30.0.17] Configuring Spark master...
HDFS online.
Spark Health Report:
  * Master: ALIVE
  * Workers: 2
  * Cores: 8
  * Memory: 57.9 GB            
launch finished in 0:02:22.
Cluster master: ec2-54-255-211-120.ap-southeast-1.compute.amazonaws.com
Login with: flintrock login test
ubuntu@ip-172-30-0-42:/flintrock_config/$ 
```

pragnesh:
[...]

PR #234, opened 2018-02-16
Set default number of SSH tries > 1

nchammas:
cc @pragnesh and @steve-drew-strong-bridge.

If we can get away with 3 tries I'd prefer that, but you let me know if 5 tries is what's needed to avoid this issue. I don't see it so I'll need your feedback here.

Prompted by #198.

### DIFF
--- a/flintrock/ssh.py
+++ b/flintrock/ssh.py
@@ -69,7 +69,9 @@ def get_ssh_client(
     if wait:
         tries = 100
     else:
-        tries = 1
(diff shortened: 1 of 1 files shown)


pragnesh:
I think 3 tries should be good enough. I will try this out and post the feedback.

pragnesh:
i have tried launching cluster multiple times with number of tries 3. So far i have not seen any issue.

nchammas:
Thank you for the feedback @pragnesh. Merging this in.

PR #242, opened 2018-03-26
Mention Homebrew install instructions

tdsmith:
Hi; thanks for flintrock! I packaged flintrock for Homebrew if you'd like to mention it in the readme. The advantage of installing with Homebrew is that it keeps flintrock and its dependencies in its own private virtualenv, and users can get new versions when they run `brew upgrade`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,8 +106,6 @@ pip3 install flintrock
 
 This will install Flintrock and place it on your path. You should be good to go now!
 
-If you use [Homebrew](https://brew.sh), you can `brew install flintrock` instead.
-
 You'll probably want to get started with the following two commands:
 
 ```sh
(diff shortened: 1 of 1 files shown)


nchammas:
Hi @tdsmith and thank you for adding Flintrock to Homebrew. I guess this is where it happened: https://github.com/Homebrew/homebrew-core/pull/25025 👍

I'm not interested in maintaining a Flintrock package for Homebrew at this time, so my concern about adding this to the README is that it "blesses" the Homebrew package as an official installation method. If the Homebrew package is not kept up to date, it will lead to some confusion and I will perhaps get some of the blame.

That said, I wouldn't mind mentioning it in the README as long as it's clear that the Homebrew package is community-maintained and that users wanting to be sure they are getting the latest version should instead stick with pip.

tdsmith:
That makes sense! Thanks for your quick feedback. I tried to address your concern by adding a separate section for community-supported distributions; what do you think?

nchammas:
Looks good to me. I'll merge this in and make some edits for wording.

Thanks again!

PR #243, opened 2018-04-02
Add links to new flintrock guides

heathermiller:
I just wrote up a guide for using flintrock to start up a Spark cluster and how to use it to submit jobs to your cluster. Have a look! Looks like it could be useful documentation for your project :)

Wonderful tool btw!

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,9 @@ Flintrock has been featured in a few talks, guides, and papers around the web.
 * Talks:
   * [Flintrock: A faster, better spark-ec2](https://www.youtube.com/watch?v=3aeIpOGrJOA) ([slides](http://www.slideshare.net/SparkSummit/flintrock-a-faster-better-sparkec2-by-nicholas-chammas))
 * Guides:
+  * **Running Spark on a Cluster: The Basics (using flintrock)**
+    * [Part 1: Start a Spark Cluster and Use the spark-shell](http://heather.miller.am/blog/launching-a-spark-cluster-part-1.html)
+    * [Part 2: Dependencies, S3, and Deploying via spark-submit](http://heather.miller.am/blog/launching-a-spark-cluster-part-2.html)    
   * [Spark with Jupyter on AWS](https://github.com/PiercingDan/spark-Jupyter-AWS)
   * [Building a data science platform for R&D, part 2 – Deploying Spark on AWS using Flintrock](https://alexioannides.com/2016/08/18/building-a-data-science-platform-for-rd-part-2-deploying-spark-on-aws-using-flintrock/)
   * [AWS EC2를 활용 스파크 클러스터 생성](http://statkclee.github.io/ml/ml-aws-ec2-flintrock.html)
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Flintrock has been featured in a few talks, guides, and papers around the web.
 * Talks:
   * [Flintrock: A faster, better spark-ec2](https://www.youtube.com/watch?v=3aeIpOGrJOA) ([slides](http://www.slideshare.net/SparkSummit/flintrock-a-faster-better-sparkec2-by-nicholas-chammas))
 * Guides:
-  * **Running Spark on a Cluster: The Basics (using flintrock)**
+  * Running Spark on a Cluster: The Basics (using Flintrock)
     * [Part 1: Start a Spark Cluster and Use the spark-shell](http://heather.miller.am/blog/launching-a-spark-cluster-part-1.html)
     * [Part 2: Dependencies, S3, and Deploying via spark-submit](http://heather.miller.am/blog/launching-a-spark-cluster-part-2.html)    
   * [Spark with Jupyter on AWS](https://github.com/PiercingDan/spark-Jupyter-AWS)


README.md
```diff
@@ -15,6 +15,9 @@ Flintrock has been featured in a few talks, guides, and papers around the web.
 * Talks:
   * [Flintrock: A faster, better spark-ec2](https://www.youtube.com/watch?v=3aeIpOGrJOA) ([slides](http://www.slideshare.net/SparkSummit/flintrock-a-faster-better-sparkec2-by-nicholas-chammas))
 * Guides:
+  * **Running Spark on a Cluster: The Basics (using flintrock)**
```
nchammas: For brevity, can we just have this line in regular font and add "(part 1 / part 2)" at the end? With "part 1" and "part 2" linking to the two parts, respectively.

Also, please capitalize Flintrock.

heathermiller:
> Could you capitalize Flintrock? The executable is indeed flintrock, but the tool and project are called Flintrock. It's similar to Docker vs. docker, or PySpark vs. pyspark.

Sure.

> In part 2 of your guide you dedicate a section to setting up dependencies for S3 access. Flintrock makes a good effort to do this for you by configuring spark.jar.packages automatically to load hadoop-aws. Did that not work for you? You should not need to install hadoop-aws-2.7.2.jar
aws-java-sdk-1.7.4.jar manually.

No, that doesn't work at all. It was quite painful to realize that Flintrock wasn't doing it when I thought it was. I don't know where Flintrock is putting the jars, but they're not available to the Spark application that I'm running when I use `spark-submit`.

README.md
heathermiller: I'd rather not because the two guides are very different.

One teaches you how to use spark-shell. One teaches you how to actually package up real (not-totally-simple) pipelines and run them on a Spark cluster. 

I think that distinction should be made. They're really two separate guides under one theme; we use Flintrock instead of spark-ec2 to do two different things.

heathermiller:
PS, just unbolded the text and capitalized Fintrock

README.md
nchammas: Understood. To clarify my thought process, the reason I ask for this change is two-fold:
1. I'm trying not to elevate any one guide above the others by granting it more real estate in such a prominent place in the README.
2. The faster this section of the README grows, the sooner I will likely move it out of the README (and likely into the Wiki). Perhaps I should have done this from the start, since it reduces the pressure on me to curate references and gives us more freedom to expand on each reference.

That said, we can leave things as you have them now but in the future I may edit or move the references section for the above reasons.

heathermiller: It's not particularly important to me if you want to "promote" what I've written or not.

I'm a professor teaching Spark to students. The documentation out there for submitting a Spark job to a cluster is really not great. For Flintrock or spark-ec2 or any other tool for that matter.

Since I've used Flintrock in my class, and since there is not much in the way of documentation for your project, I merely offer the guide to you in case you would like to use it in place of maintaining documentation yourself.

How and whether you manage your project is up to you! Just note that now that spark-ec2 is deprecated, I imagine you're going to have many more people looking for documentation/guides/etc on setting up such pipelines...

nchammas:
> It was quite painful to realize that Flintrock wasn't doing it when I thought it was. I don't know where Flintrock is putting the jars, but they're not available to the Spark application that I'm running when I use spark-submit.

We should probably fork this discussion to a dedicated issue, but to elaborate briefly here:

Flintrock configures a default for `spark.jar.packages` that `spark-submit` picks up automatically. I _think_ the reason it didn't work for you, looking at your guide, is that you call `spark-submit` from your workstation rather than from the cluster. Or maybe it's that you [used the cluster deploy mode](https://github.com/nchammas/flintrock/pull/180#issuecomment-276037851)? I'm not sure.

If you login to the cluster and just call `pyspark` or `spark-shell` from there you should see Spark loading all the `hadoop-aws` dependencies. It's the same as if you had called `pyspark`, `spark-shell`, or `spark-submit` with `--packages "org.apache.hadoop:hadoop-aws:2.7.5"`. You can see more background on how we built this feature over on #180.

I'd like to fix this so that Flintrock does the right thing for S3 dependencies regardless of how the user submits applications. I'll file an issue for this. Do you mind if I pinged you from there for feedback?

README.md
nchammas: I 100% appreciate the guides that you and others have written and believe they help close the gap between having a tool that works and having a tool that people feel comfortable using. I'm happy to link to them from the project and think they are helpful.

I'm just thinking out loud as a maintainer about how I'd like to not be in the business of curating external references. It's not about any individual guide. So for example, the note about wanting the reference to be limited to one line was motivated by a hypothetical future scenario where another guide author messages me, upset, that I've given guide X more prominence than their guide Y. And the note about potentially moving references to the Wiki was motivated by another hypothetical where more and more guide authors want to add links to their work, making the README a bit unwieldy.

It's silly, but that's the kind of stuff I worry about. 😄 I hope that clarifies my intentions. I hadn't really thought through how to manage the references section (which I only recently added to the README), so this PR happened to provide me the opportunity to hash my thought process out. Please accept my apologies for turning a simple PR into a lengthy discussion. 😅

heathermiller:
That would be great!

And sure thing! Happy to test it out!

heathermiller:
Oh, and since I was setting this guide up for students... It should be quite reproducible!

So if you just use the little example program in part 2 guide using spark-submit, and if you drop the part about manually copying the s3 jars to the cluster with Flintrock, you should see the typical NoClassDef found or whatever exception that's thrown when trying to use one of the S3 methods.

So that's at least one nice thing. Hopefully a reproducible example for this issue 😛

Issue #236, opened 2018-02-16
Flintrock Describe errors if one of the clusters is shutting down


steve-drew-strong-bridge:
* Flintrock version:  0.9
* Python version:  3.4.7
* OS:  aws linux

When trying to list the clusters that are still running, I receive an untrapped error when one of the clusters is in the process of shutting down.  Running this same command a minute later will work, as 

Expected: continue to describe the instances and maybe give a brief eulogy for the dead cluster.

flintrock describe --ec2-region=us-west-2

Traceback (most recent call last):
  File "/usr/local/bin/flintrock", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib64/python3.4/site-packages/flintrock/flintrock.py", line 1132, in main
    cli(obj={})
  File "/usr/local/lib/python3.4/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.4/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.4/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.4/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.4/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.4/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/usr/local/lib64/python3.4/site-packages/flintrock/flintrock.py", line 510, in describe
    vpc_id=ec2_vpc_id)
  File "/usr/local/lib64/python3.4/site-packages/flintrock/ec2.py", line 1009, in get_clusters
    _get_cluster_name(instance) for instance in all_clusters_instances}
  File "/usr/local/lib64/python3.4/site-packages/flintrock/ec2.py", line 1009, in <setcomp>
    _get_cluster_name(instance) for instance in all_clusters_instances}
  File "/usr/local/lib64/python3.4/site-packages/flintrock/ec2.py", line 1063, in _get_cluster_name
    i=instance.id))
Exception: Could not extract cluster name from instance: i-0f4f75fb5f429994f

dm-tran:
[...]

PR #246, opened 2018-04-11
Fix some "flintrock describe" errors, using a filter on instance-state-name

dm-tran:
This PR makes the following change:
* Add a filter to ignore instances that are 'shutting-down' or 'terminated'. (actually, we can't exclude states, so we have to specify the allowed states, i.e. 'pending', 'running', 'stopping' and 'stopped')

I tested this PR by
* launching a cluster
* adding slaves
* stopping the cluster
* removing slaves
* destroying the cluster.

For each tested command, I ran "flintrock describe" using two standalone Flintrocks (one with the fix and one without).

This PR partially fixes `Exception: Could not extract cluster name from instance: when a cluster is destroyed or when slaves are removed` (#236): this exception will be raised during the first seconds instead of being raised during 30 seconds / a minute.

This PR also fixes the following error, which is raised when a cluster is launched or when slaves are added:
```
  File "standalone.py", line 11, in <module>
  File "flintrock/flintrock.py", line 1110, in main
  File "click/core.py", line 716, in __call__
  File "click/core.py", line 696, in main
  File "click/core.py", line 1060, in invoke
  File "click/core.py", line 889, in invoke
  File "click/core.py", line 534, in invoke
  File "click/decorators.py", line 17, in new_func
  File "flintrock/flintrock.py", line 477, in describe
  File "flintrock/ec2.py", line 1053, in get_clusters
  File "flintrock/ec2.py", line 1053, in <listcomp>
  File "flintrock/ec2.py", line 1104, in _compose_cluster 
  File "flintrock/ec2.py", line 1080, in _get_cluster_master_slaves
TypeError: 'NoneType' object is not iterable
```
Instead of raising this error, `flintrock describe` will now display :
```
my-cluster-name:
  state: inconsistent
  node-count: 0
```

### DIFF
--- a/flintrock/ec2.py
+++ b/flintrock/ec2.py
@@ -1002,6 +1002,7 @@ def get_clusters(*, cluster_names: list=[], region: str, vpc_id: str) -> list:
     all_clusters_instances = list(
         ec2.instances.filter(
             Filters=[
+                {'Name': 'instance-state-name', 'Values': ['pending', 'running', 'stopping', 'stopped']},
                 {'Name': 'instance.group-name', 'Values': group_name_filter},
                 {'Name': 'vpc-id', 'Values': [vpc_id]},
             ]))


nchammas:
Thank you for submitting this @dm-tran! Your changes look good to me.

> This PR also fixes the following error, which is raised when a cluster is launched or when slaves are added:

To be clear, do you mean this error is raised if you call `flintrock describe` while another instance of Flintrock is launching or expanding a cluster? Or do you mean that `flintrock launch` is sometimes throwing this error directly?

nchammas:
OK, thanks for clarifying.

PR #252, opened 2018-05-29
Track transitive dependencies via pip-tools

### LINKED ISSUE #248: Make standalone builds deterministic
If you are reporting a potential bug, include as many of the following
as are relevant (and delete this sentence, of course):
* Flintrock version: 0.9
* Python version: any
* OS: any

Dependencies for standalone builds should be deterministic, because the build is completely isolated from any other installed python package.
[Pipenv](https://github.com/pypa/pipenv) or [poetry](https://github.com/sdispater/poetry) could be a good solution to this problem.

Also, the build environment (Docker?) should be generalized and documented.

nchammas:
I think in parallel to this I will try to get Poetry working. Just wanted to get a feel for this approach.

Advantages to using pip-tools:
* Approach is conservative. Just adds a relatively simple tool to the toolchain.
* Behavior is easy to understand.

Disadvantages:
* Adds a lot of new files to track.
* ~Does not handle `-e .` well. (See https://github.com/jazzband/pip-tools/issues/663.)~ There is a decent [work-around](https://github.com/nchammas/flintrock/pull/252/files#diff-bc49d3325564a726d5f5777378dbdc1aR7-R9) for this.

@Perados, what do you think?

This doesn't simplify the problem of packaging. I'm also not sure if it fully addresses the issue you raised, since the compiled dependencies don't affect the package that gets published to PyPI. This may be OK, though, since you were interested in building your own packages reliably. And if the other approaches out there like Poetry and Pipenv prove to be a lot more work, then at least we have something here we can use in the interim.

Fixes #248.


nchammas:
Tested this end-to-end with a full run of the test suite against AWS.

PR #260, opened 2018-07-14
xfail flaky get_latest_commit() test on Travis

nchammas:


nchammas:
Annoyingly, I cannot get this test to XFAIL on Travis CI to confirm that I've fixed the issue, even after a dozen or so test builds. The test is now consistently passing. So I'm going to merge this in and potentially follow-up with another fix if/when the problem actually happens again.

PR #262, opened 2018-07-17
[WIP] Migrate packaging to Poetry

### LINKED ISSUE #248: Make standalone builds deterministic

nchammas:
Alternative to #252.

- [x] Update Travis CI build
- [x] Understand impact on PyInstaller distributions
    - Seems to be no impact
- [ ] Test publishing of sdist and wheel
    - [ ] Are all the same files included?
- [ ] Do we still need `__main__.py`?
- [ ] Update README
- [ ] Update Contributing guide

Potentially blocking issues:
* ~Getting `flintrock` on the `PATH`: https://github.com/sdispater/poetry/issues/311~
* Dealing with multiple Python versions: https://github.com/sdispater/poetry/issues/354

Other issues:
* No clean way to single-source the application version: https://github.com/sdispater/poetry/issues/273

Fixes #248.


nchammas:
I think this might be worth revisiting some other time. For now, I'll stick with something more conservative, like in #252.

PR #266, opened 2018-08-06
Change the config file doc

dirceusemighini:
Adding the --config to the launch with config file that I think it was missing.

* 
* 

I tested this PR by...

Fixes #NN.
Fixes #MM.

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,7 +272,7 @@ providers:
 With a config file like that, you can now launch a cluster with just this:
 
 ```sh
-flintrock launch test-cluster
+flintrock --config config.yaml launch test-cluster
 ```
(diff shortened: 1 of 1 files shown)


nchammas:
Hi @dirceusemighini. If you call `flintrock configure` and edit the file that gets opened, you won't need to pass `--config`.

Only if you have a config file in a non-default location do you need to specify `--config`.

dirceusemighini:
Sorry @nchammas by some reason the configure didn't worked in my machine and I took sometime to realize that I wasn't using the config.yaml file.
I thought that only creating a config.yaml file would make the flint to use it

PR #258, opened 2018-07-12
Add Python 3.7 to Travis CI build

nchammas:
The required PyYAML upgrade was taken care of in #268.

PR #268, opened 2018-08-07
Updated PyYaml dependency

### LINKED ISSUE #267: PyYaml dependency (3.12) incompatible with Python 3.7
* Flintrock version: 0.8.0
* Python version: 3.7.0
* OS:  Arch Linux (4.17.11-arch1 x86_64)

I don't know if you are aware of this, but the latest version of flintrock fails to build with Python 3.7.0
I was experimenting with it yesterday and after I was finished I carelessly launched a system upgrade; apparently PyYaml 3.12 (the version currently used by flintrock) is not compatible with Python 3.7 (referencing [this issue](https://github.com/yaml/pyyaml/issues/126)), and I had to fall back to 3.6 to work with flintrock again.

PyYaml 3.13 on the opposite works with the newest version. I fixed my problem, but maybe it would be best to update the PyYaml dependency from 3.12, if there are no inherent problems with that?

Maldus512:
bumped up PyYaml version to version 3.13, which works under Python 3.7.0

- Simply updating the PyYaml dependency in setup.py

I tested this PR by installing the resulting package and running some basic usage (launch, destroy, login, copy-file).

Fixes #267.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,7 +53,7 @@ setuptools.setup(
         'botocore == 1.10.58',
         'click == 6.7',
         'paramiko == 2.4.1',
-        'PyYAML == 3.12',
+        'PyYAML == 3.13',
         # This is to ensure that PyInstaller works. dateutil is an
         # indirect dependency of Flintrock, and PyInstaller chokes on
         # dateutil 2.5.0.


nchammas:
This PR doesn't fix #258 because we still want to separately add Python 3.7 to the Travis CI build. Please update the PR description accordingly.

nchammas:
Thanks @Maldus512!

PR #277, opened 2018-11-01
[#271] Avoid multiple client IP addresses by using HTTPS

### LINKED ISSUE #271: Flintrock throws an exception when "checkip" returns multiple IP addresses
* Flintrock version: 0.10.0
* Python version: 3.6.4
* OS: Debian

For the network I'm in, "checkip" returns multiple IP addresses separated by commas. Flintrock just adds a subnet mask and tries to create the security group rules. Then I get the following exception:

```python
botocore.exceptions.ClientError: An error occurred (InvalidParameterValue) when calling the AuthorizeSecurityGroupIngress operation: CI$
R block <IP addresses> is malformed
```

Line where the exception is thrown: https://github.com/nchammas/flintrock/blob/master/flintrock/ec2.py#L600

Might a simple solution be just splitting the string and using only the first IP?

boechat107:
* replaces HTTP protocol by HTTPS when calling AWS service **checkip**

I tested this PR by calling the service from many different clients using both HTTP and HTTPS protocols.

As suggested in #272, just replacing the protocol seems to avoid the problem of receiving multiple IP addresses from **checkip**.

Fixes #271

### DIFF
--- a/flintrock/ec2.py
+++ b/flintrock/ec2.py
@@ -541,7 +541,7 @@ def get_or_create_flintrock_security_groups(
 
     # Rules for the client interacting with the cluster.
     flintrock_client_ip = (
-        urllib.request.urlopen('http://checkip.amazonaws.com/')
+        urllib.request.urlopen('https://checkip.amazonaws.com/')
         .read().decode('utf-8').strip())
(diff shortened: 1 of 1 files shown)


nchammas:
Thanks for sticking through the research on this issue @boechat107 and settling on this nice, simple solution!

PR #279, opened 2018-11-14
Simplify Python 3.7 portion of Travis CI script

nchammas:
https://github.com/travis-ci/travis-ci/issues/981 has been resolved, so perhaps the script can be made simpler now.

Relates to #258.


nchammas:
Looks like we might need to switch to Xenial for the whole build. Not interested or needed at this time, so I'll close this PR and revisit it in the future.

PR #285, opened 2019-01-25
switch all worker to private dns

pragnesh:
This PR makes the following changes:

switch all worker to private dns

I tested this PR by...
launching cluster where i am seeing slow startup(flintrock needed bend-aid to start cluster).
After this change cluster start in first try.

Fixes #157.
Fixes #129.

When instance start on aws, aws update public dns host from public ip -> private ip,
This create issue with cluster communication. If you switch all worker to private dns name,
Cluster does not have any issue in starting which fix band-aid required for starting hdfs cluster and spark cluster.

PR #296, opened 2019-12-16
Add support to private_vpc (ie: with no public address)

### LINKED ISSUE #14: Support launching clusters into private VPCs

luhhujbb:
* support of private_vpc (ie: with public ip address like this : https://github.com/chaordic/flintrock/pull/4/files)
* add client-source option since in private mode, the IP returned by checkip.amazonaws.com will be a nonsense in most of cases (using a VPN, for instance) and it's sounds good to have all good ports configured automatically in this case, even if custom security groups can be added. 
* client-source support both cidr address or security group.

Default behaviour remains identical except that it logs when flintrock detect private mode instead of raising an exception.
Fixes #14


nchammas:
Hey @luhhujbb, thanks for submitting this PR. I will take a look at it over the holidays!

flintrock/ec2.py
```diff
@@ -310,12 +325,12 @@ def add_slaves(
                     ])
                 .create_tags(Tags=slave_tags))
 
-            existing_slaves = {i.public_ip_address for i in self.slave_instances}
+            existing_slaves = {i.public_ip_address or i.private_ip_address for i in self.slave_instances}
```
nchammas: It seems like if we have a new property to explicitly detect if we're running in a private subnet, `self.private_net`, we should use it here too (and just below) to switch between public and private addresses.

flintrock/ec2.py
```diff
@@ -493,7 +507,8 @@ def get_or_create_flintrock_security_groups(
         *,
         cluster_name,
         vpc_id,
-        region) -> "List[boto3.resource('ec2').SecurityGroup]":
+        region,
+        client_source) -> "List[boto3.resource('ec2').SecurityGroup]":
```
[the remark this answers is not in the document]
luhhujbb: * client-source doesn't add arbitrary rules, a user in a private environment doesn't want to think about different ports which are already defined in flintrock. One only wants basic rules to be setup where the flintrock client run. 
* client-source also disable public ip security rules configuration which is on a private environment quite useful in terms of security.

flintrock/ec2.py
```diff
@@ -71,6 +71,9 @@ def __init__(
         self.vpc_id = vpc_id
         self.master_instance = master_instance
         self.slave_instances = slave_instances
+        # check private mode
+        ec2 = boto3.resource(service_name='ec2', region_name=self.region)
+        self.private_net = not ec2.Subnet(self.master_instance.subnet_id).map_public_ip_on_launch
```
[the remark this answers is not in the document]
luhhujbb: Maybe I'm wrong, but a @property decorator implies that the the following code will be executed each times the property is get, that is not efficient. This way it reduces aws api call.

luhhujbb:
Test passes with USE_AWS_CREDENTIALS=true in the following condition:
* public subnet (default behaviour)
* private subnet, with client-source option configured with my vpn IP.

flintrock/ec2.py
luhhujbb: Fixed.

flintrock/ec2.py
```diff
@@ -71,6 +71,9 @@ def __init__(
         self.vpc_id = vpc_id
         self.master_instance = master_instance
         self.slave_instances = slave_instances
+        # check private mode
+        ec2 = boto3.resource(service_name='ec2', region_name=self.region)
+        self.__private_net = not ec2.Subnet(self.master_instance.subnet_id).map_public_ip_on_launch
```
nchammas: I don't think a double underscore variable name is appropriate here. Those are for special methods and "class-private" members, which is not the case here.

flintrock/ec2.py
```diff
@@ -79,21 +82,37 @@ def instances(self):
         else:
             return self.slave_instances
 
+    @property
```
nchammas: My intention in recommending a `@property` was that all the logic would be contained within it. Yes, there is a penalty in that the query to AWS will be made (ignoring any caching that boto3 may do) every time you access the property, but that minor downside is outweighed in my view by the clearer API. The property doesn't get called that often.

So I'd say to either put everything in the property method, or go back to having it just be a plain old property like you did before. It's not a big deal either way.

tests/conftest.py
```diff
@@ -12,7 +12,7 @@
 
 HADOOP_VERSION = '2.8.5'
 SPARK_VERSION = '2.4.4'
-SPARK_GIT_COMMIT = 'c3e32bf06c35ba2580d46150923abfa795b4446a'  # 2.4.3
+SPARK_GIT_COMMIT = '7955b3962ac46b89564e0613db7bea98a1478bf2'  # 2.4.4
```
nchammas: Did you need to make this change to get tests to pass? I'd keep it out of the PR since it's unrelated to the topic at hand.

flintrock/flintrock.py
```diff
@@ -316,6 +316,7 @@ def cli(cli_context, config, provider, debug):
               multiple=True,
               help="Additional tags (e.g. 'Key,Value') to assign to the instances. "
                    "You can specify this option multiple times.")
+@click.option('--client-source', type=str, default=None)
```
nchammas: I think we need to improve a few things here.

1. I'd rename this option to `--ec2-authorize-access-from`. It's clearer, and it categorizes the option under EC2 where it belongs.
2. I'd also add a brief help text explaining what this option does and what input it takes, since it may not be obvious to everyone. Since this option disables the automatic detection of the client IP, we should call that out too.
3. We should somehow validate the input here since we are expecting either a CIDR or a security group ID. A user might, as I did, assume they can provide a plain IP address, which then leads to the following error:

    ```
    botocore.exceptions.ClientError: An error occurred (InvalidParameterValue) when calling the 
    AuthorizeSecurityGroupIngress operation: CIDR block 172.31.38.197 is malformed
    ```

    Either that, or we should accept a plain IP address and automatically covert it into a `/32` CIDR block.

    Take a look at `--ec2-tag`  for an example of how to validate user input.
4. Since we are adding a new option, we need to add it to the [config template](https://github.com/nchammas/flintrock/blob/master/flintrock/config.yaml.template).

nchammas:
This is looking good @luhhujbb. I will test this out tomorrow and try to get a private VPC working in my environment. I think I need to setup a [NAT gateway](https://docs.aws.amazon.com/vpc/latest/userguide/vpc-nat-gateway.html) to enable the private VPC to still talk to the outside world.

Do you mind if I push some commits to your branch? It will mostly be to tweak some method names or message text that is sent to the user. I've been very slow with the review feedback, so for small things I'd like to spare you the time and just do them myself.

If you're OK with that, let me know and just make sure that the "Allow edits from maintainers" checkbox is enabled in the right side-bar.

luhhujbb:
Hi, @nchammas, of course you can edit, the box "Allow edits from maintainers" is checked.

nchammas:
I've successfully created a private subnet that accesses the Internet via a NAT gateway on a public subnet in the same VPC. I've also successfully tested out this PR by launching a cluster in that environment. So far so good. 👍

I've pushed the latest changes from master onto your fork, @luhhujbb, and my next steps will be to retest your PR given the recent changes introduced by #285. I will also push some edits as noted in my previous comment.

(By the way, for the future you may want to make contributions from a branch on your fork, as opposed to from master. It will make pulling updates from the upstream source easier down the line.)

kedar700:
Any update on when this might get merged

nchammas:
Hello @kedar700. Sorry, I haven't been able to look at this PR recently.

I think this PR has the right implementation but, before it can be merged in, it needs some tweaks and needs merge conflicts to be resolved.

It's my fault for leaving this PR in limbo for so long. (Sorry @luhhujbb!)

luhhujbb:
I've rebase from master. The python 3.5 compatibility was difficult to handle. Maybe it should be dropped since python 3.5 reaches its end of life.
 
`authorize-access-from` is now an array of configurable security-group/network cidr, it allows flexible configuration in a private environment.

nchammas:
Thank you for updating this PR @luhhujbb. I've dropped Python 3.5 support in #329 and will take a look at your updates here later this week. Hopefully we can get this merged in this time around!

luhhujbb:
@nchammas I've just rebase. Waiting for your review. Thanks for your time.

flintrock/core.py
```diff
@@ -76,6 +76,16 @@ def master_host(self) -> str:
         """
         raise NotImplementedError
 
+    @property
+    def private_net(self) -> bool:
```
nchammas, line 80: Nit: Could we rename this `private_network()`?

flintrock/ec2.py
```diff
@@ -79,25 +80,43 @@ def instances(self):
...
+    def private_net(self):
+        # check private mode
```
nchammas, line 85: I'd delete this comment. It's unnecessary.

flintrock/flintrock.py
```diff
@@ -319,6 +319,17 @@ def cli(cli_context, config, provider, debug):
...
+              callback=ec2.cli_validate_ec2_authorize_access,
+              type=str,
+              default=None,
+              multiple=True,
+              help="Authorize cluster access from a private source"
+                   "(To be used in VPC context, behind a nat gateway for instance)"
+                   "which can be either a plain IP,"
+                   "an IP network in CIDR notation, or an EC2 SecurityGroup."
+                   "For security concerns, this options disable client configuration "
+                   "using automatic detection of client public IP")
```
nchammas, line 332: I forget if we already discussed this, but does the `add-slaves` command automatically respect this setting? I assume yes, since this is done at the level of the cluster security group, but just confirming.

nchammas (commented):
Took a quick look now. Looks good. Will take a closer look and retest later this week.

flintrock/ec2.py
luhhujbb: Ok

flintrock/flintrock.py
luhhujbb: Slaves get their security groups from cluster master ones, so it's not needed:

https://github.com/nchammas/flintrock/blob/d3f769e40d1cce742996412b6a2dc2956870fc93/flintrock/ec2.py#L258-L260

nchammas:
@luhhujbb - I made some changes to the PR to tweak the wording or style in various places, and also to refactor how we are validating the CLI input. I also think I fixed some bugs, like `.appends -> .append`. I tested my changes by launching a cluster into a private VPC with both Hadoop and Spark installed.

1. What do you think of the latest changes?
2. Can you test and confirm that the current state of the PR still works for your private VPC use case?

nchammas (approved):

luhhujbb:
@nchammas First of all many thanks for your time !
1. I've check all your commits and it looks good.
2. I've successfully launch a production cluster using this version of flintrock. Everythinks is Ok. Using multipe "ec2-autorize-access-from" allows us to meet all our production use cases.

nchammas:
Thanks for persisting through the lengthy review/refactor process! I'm glad we got this in.

PR #304, opened 2020-04-01
Properly decode UserData before adding to new slaves

### LINKED ISSUE #303: User data scripts aren't executed when adding slaves
I _think_ I've found a bug in the `add-slaves` command (or maybe I'm just misunderstanding how it's supposed to work). Here are my version specs:

* Flintrock version: 1.0.0
* Python version:  3.7.4
* OS: MacOS Catalina 10.15.4 (19E266)

I create my cluster with a configuration file that looks like this:
```
services:
  spark:
    version: 2.4.5
  hdfs:
    version: 2.7.7

provider: ec2

providers:
  ec2:
    ...  # Omitting region, availability zone, etc here for brevity
    instance-type: m5.large
    ami: ami-0ce49ca477b768354
    user-data: scripts/install-python3.sh

launch:
  num-slaves: 2
  install-hdfs: True
  install-spark: True

debug: false
```

The user data script at `scripts/install-python3.sh` installs Python 3.7 and looks like this:
```
#!/usr/bin/env bash

set -e

sudo yum install -y python3 python3-devel python3-pip python3-setuptools python3-virtualenv
```

As expected, when I run `flintrock launch` the cluster is created and each node has Python 3 installed. I've verified this manually by SSH'ing into each node and checking.

Next, I add a single node by running `flintrock add-slaves --num-slaves 1`. When I log into the new node, Python 3.7 is unavailable. This is an unexpected result (in my opinion), but I'm not sure whether it's the intended behaviour or not so I'm reporting it here.

Thanks for the otherwise excellent tool! :)

nchammas:
I tested this PR by adding slaves to an existing cluster and confirming that the `UserData` attribute is not gibberish.

Fixes #303.

### DIFF
--- a/flintrock/ec2.py
+++ b/flintrock/ec2.py
@@ -278,7 +278,10 @@ class EC2Cluster(FlintrockCluster):
         if not response['UserData']:
             user_data = ''
         else:
-            user_data = response['UserData']['Value']
+            user_data = (
+                base64.b64decode(response['UserData']['Value'])
+                .decode('utf-8')
+            )
 
         if not self.master_instance.iam_instance_profile:
             instance_profile_arn = ''


nchammas:
@dhorgan - Test this out and let me know if it fixes the problem for you.

I'm curious to know also if you have any issues with `UserData` when using spot instances.

dhorgan:
I've just tested this and I can confirm that it's working now! 👍

dhorgan:
> I'm curious to know also if you have any issues with `UserData` when using spot instances.

Sorry, I can't help here - we aren't using spot instances right now.

nchammas:
No worries. Thanks for confirming on your end.

I tested out spot instances just now, and adding a spot slave preserves the `UserData` attribute correctly.

Merging this in.

PR #307, opened 2020-05-06
Hadoop 3.x port changes (#306)

### LINKED ISSUE #306: Add support for HDFS 3.x+
Flintrock works great with the new 3.0 Apache Spark releases.

Unfortunately we cannot bring along Hadoop 3.x because they have changed many of their standard TCP port numbers. See [Default ports of multiple services have been changed.](https://hadoop.apache.org/docs/r3.0.0/index.html#Default_ports_of_multiple_services_have_been_changed.)

From the Flintrock perspective, port 9870 must be used in place of port 50070.

sfcoy:
 * Migrate SecurityGroupRules provisioning to their respective FlintrockService classes (as per the TODO in the source)
 * Use the HDFS version to select the name node ui port number

I tested this PR by creating and launching the following combinations of Spark and Hadoop:
 * Spark 2.4.5, Hadoop 2.8.5
 * Spark 3.0.0rc1. Hadoop 3.2.1

Fixes #306

### DIFF
--- a/flintrock/services.py
+++ b/flintrock/services.py
@@ -115,9 +115,9 @@ class FlintrockService:
         """
         raise NotImplementedError
 
-    def create_security_group_rules(self, flintrock_client_cidr: str):
+    def get_security_group_rules(self, flintrock_client_cidr: str):
         """
-        Create and return the EC2 SecurityGroupRules required by this service.
+        Return the EC2 SecurityGroupRules required by this service.
         """
         raise NotImplementedError
 
@@ -256,7 +256,7 @@ class HDFS(FlintrockService):
         except Exception as e:
             raise Exception("HDFS health check failed.") from e
 
-    def create_security_group_rules(self, flintrock_client_cidr: str):
+    def get_security_group_rules(self, flintrock_client_cidr: str):
         return [
             SecurityGroupRule(
                 ip_protocol='tcp',
@@ -450,7 +450,7 @@ class Spark(FlintrockService):
             #       dump a large stack trace on the user.
             raise Exception("Spark health check failed.") from e
 
-    def create_security_group_rules(self, flintrock_client_cidr: str):
+    def get_security_group_rules(self, flintrock_client_cidr: str):
         return [
             SecurityGroupRule(
                 ip_protocol='tcp',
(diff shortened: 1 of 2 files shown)


nchammas:
Hello @sfcoy and thank you for this PR. I still have this on my list to review. Sorry about the slow turnaround time! I will get back to you in the next week or so.

flintrock/services.py
```diff
@@ -106,11 +115,18 @@ def health_check(
         """
         raise NotImplementedError
 
+    def create_security_groups(self, flintrock_client_cidr: str):
+        """
+        Create and return the EC2 SecurityGroupRules required by this service
```
nchammas, line 120: [Nit](https://stackoverflow.com/a/27810540/877069): Full stop at the end of the sentence, please.

flintrock/services.py
```diff
@@ -106,11 +115,18 @@ def health_check(
...
+        """
+        return []
```
nchammas, line 122: Why `return []` and not `raise NotImplementedError`? I would prefer the latter for consistency, unless there is some reason to do otherwise.

flintrock/services.py
```diff
@@ -240,6 +256,16 @@ def health_check(self, master_host: str):
         except Exception as e:
             raise Exception("HDFS health check failed.") from e
 
+    def create_security_groups(self, flintrock_client_cidr: str):
```
nchammas, line 259: `create_security_groups()` suggests this method will actually create the groups, which isn't what it does. I think `get_security_groups()` would be a better name.

flintrock/services.py
sfcoy: I was torn between the two at the time to be honest. I've changed it as requested.

flintrock/services.py
sfcoy: I will rename to `create_security_group_rules` if you don't mind. Personally I find the use of `get...` to be overused in function names, especially coming from Java land.

nchammas: It's just that "create" suggests the rules will actually be deployed to AWS, but this method only produces a list of rules that then get created/deployed later in ec2.py. Hence the suggestion to use "get" or some other verb that doesn't imply actual creation of the resource. But it's not a big deal.

sfcoy:
Mate, how could I possibly mind? It's your baby and we are grateful to have access to it.

nchammas:
Thank you. I am likewise grateful to have positive users and helpful and empathetic contributors like yourself.

PR #315, opened 2020-09-17
Allow configuration of EC2 spot request duration

mblack20:
When spinning up ec2 spot instances with flintrock for ad-hoc analysis, there can be cases where capacity is not available for a given instance type. Of course, one can wait until AWS releases instances, however often I find that I just cancel the request and spin up a different type of instance. However, as the default timeout of spot requests is 7 days, it may be that after several minutes/hours/days AWS finally release capacity for a cancelled / unneeded request. This then leads the situation where there are "inconsistent" clusters spun up in AWS, and flintrock has not fully configured them. In this case the instances must be manually terminated from the EC2 console. 

Additionally, there can be the scenario where a request is cancelled, a new request for a different instance is then created (with the same name), and both requests end up being fulfilled, which the causes issues with describe/destroy. In this PR, I have added support for using a human readable timedelta (e.g. `'7d'`, `'5m'`) in configuration/cli such that the spot request timeout can be controlled by the user (and it retains the default of 7 days). This would alleviate the problems described above, where I could simply reduce the spot request duration to be much shorter than 7 days.

* Adds support for the `ValidUntil` parameter when creating ec2 spot requests
* Retains the default option where requests are valid for 7 days

I tested this PR by running pytest locally and adding new tests for the util function. However, I have not run any of the acceptance tests.

### DIFF
--- a/flintrock/ec2.py
+++ b/flintrock/ec2.py
@@ -798,7 +798,7 @@ def launch(
         user,
         security_groups,
         spot_price=None,
-        spot_request_valid_until=None,
+        spot_request_duration=None,
         min_root_ebs_size_gb,
         vpc_id,
         subnet_id,
@@ -868,10 +868,10 @@ def launch(
     else:
         user_data = ''
 
-    if spot_request_valid_until is None:
+    if spot_request_duration is None:
         spot_request_valid_until = datetime.utcnow() + timedelta(days=7)
     else:
-        spot_request_valid_until = datetime.utcnow() + duration_to_timedelta(spot_request_valid_until)
+        spot_request_valid_until = datetime.utcnow() + duration_to_timedelta(spot_request_duration)
 
     try:
         cluster_instances = _create_instances(
--- a/flintrock/ec2.py
+++ b/flintrock/ec2.py
@@ -5,7 +5,7 @@ import time
 import urllib.request
 import base64
 import logging
-from datetime import datetime
+from datetime import datetime, timedelta
 
 # External modules
 import boto3
@@ -25,6 +25,7 @@ from .exceptions import (
 )
 from .ssh import generate_ssh_key_pair
 from .services import SecurityGroupRule
+from .util import duration_to_timedelta
 
 logger = logging.getLogger('flintrock.ec2')
 
@@ -653,6 +654,7 @@ def _create_instances(
         num_instances,
         region,
         spot_price,
+        spot_request_valid_until,
         ami,
         assume_yes,
         key_name,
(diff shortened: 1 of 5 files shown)


nchammas:
Hello @mblack20 and my apologies about taking so long to chime in here! Thank you for contributing to Flintrock. I will take a look at your work and give some feedback this week.

flintrock/ec2.py
```diff
@@ -864,11 +868,17 @@ def launch(
     else:
         user_data = ''
 
+    if spot_request_duration is None:
+        spot_request_valid_until = datetime.utcnow() + timedelta(days=7)
+    else:
+        spot_request_valid_until = datetime.utcnow() + duration_to_timedelta(spot_request_duration)
```
nchammas, line 874: Instead of `utcnow()`, the Python docs recommend using a timezone-aware method like `now()`.

For example:

```python
>>> from datetime import datetime, timezone
>>> datetime.now(tz=timezone.utc)
datetime.datetime(2020, 10, 18, 22, 38, 26, 703948, tzinfo=datetime.timezone.utc)
```

nchammas (commented):
This is clean and well-written. Nice work.

I would like to avoid adding more options for Flintrock, especially options that just mirror underlying EC2 options.

This is mostly because because of a design deficiency in Flintrock. I didn't think of a way to elegantly and automatically expose all of EC2's underlying options in Flintrock, so we have to manually add new options one by one to expose the underlying functionality. The more we add, the larger and messier Flintrock's interface becomes.

If this option is useful to you and you use Flintrock, then I would consider accepting this PR. But before doing that, I have to explore one alternative:

Is there a better default Flintrock can set for the spot request duration? One that would be good enough for most people. We could just set a lower default and not expose any new options.

Do you think that would be good enough?

mblack20:
Thanks for the review!

I can understand the reticence to add _yet another_ parameter, however I would argue this is a useful one, as it can help out with issues caused by late fulfilment of clusters, duplicate instances with the same name etc., and in fact in 2+ years of production use of flintrock this has been the only parameter I've been missing.

As for lowering the defaults, that's definitely the easier option, but I'm not sure what a sensible default would be. For example, I'd set it as low as 5 minutes, as if the instance type isn't available immediately I'd rather relaunch with another type. However, I'm not sure this would be suitable for all users and is a big departure from the default of 7 days, hence I opted for moving it to configuration instead.

Happy to hear your thoughts. Thanks!

nchammas (approved):
Okie doke, makes sense to me. Thanks for the feedback, for the well-written PR, and for the tests!

I'm curious to learn more about how you've been using Flintrock in production, just for my own sake. It's always an honor to hear that Flintrock has been useful to people, and I'm still impressed after so many years to hear that people are using it in production.

If you have a minute, please share a bit about what you're doing with Flintrock, either here or via my email address!

nchammas:
Looks like we missed adding this new option to the `add-slaves` command.

@mblack20 - No worries if you've moved on, but if you have the time, do you mind adding this new option to the `add-slaves` command, and testing it manually?

mblack20:
Ah sorry I missed that, yes I'll take a look and follow up with a PR

Issue #313, opened 2020-07-20
Spark 3.0.0 incompatibilities


Wadaboa:
Hi,
I'm using Flintrock to deploy a Scala/Spark application to a simple EC2 cluster (master + 1 slave), but I'm having some problems. In particular, I'm using the following custom Flintrock configuration to launch the cluster:
```yaml
provider: ec2

services:
  spark:
    version: 3.0.0

launch:
  num-slaves: 1
  install-spark: True

providers:
  ec2:
    key-name: my-key-pair
    identity-file: my-key-pair.pem
    instance-type: m4.large
    region: us-east-1
    ami: ami-00b882ac5193044e4
    user: ec2-user
    instance-profile-name: ec2-s3

debug: true
```

As you can see, I'm using Spark 3.0.0. Once I have packaged my Scala application (using `sbt`) I perform a Flintrock's `copy-file` operation. Then, when I want to run the JAR using `spark-submit`, I have issues:
* If I run `spark-submit` from my local machine using option `--master spark://<ec2-master-name>:7077` and `--deploy-mode cluster`, I get a Java error saying `Cannot run program "/usr/local/opt/openjdk@11/bin/java" (in directory "/home/ec2-user/spark/work/driver-20200718133018-0005")`
* If I run `spark-submit` from inside the master cluster (using either Flintrock's `run-command --master-only` or `login` commands), the job gets correctly submitted, but the spark UI is not reporting it (at `<ec2-master-name>:8080`). Instead, the application execution can be viewed at `<ec2-master-name>:4040`, even though in the `Executors` page only the master seems to be active.

I can't wrap my head around this problem. I don't know if it is some kind of Flintrock's compatibility issue with Spark 3.0.0, or I'm doing something wrong.
I apologize if this is the wrong place where to post this kind of doubts.

However, my environment is the following:
* Flintrock version: `1.0.0`
* Python version: `3.8.4`
* OS: `macOS Catalina 10.15.5`
* SBT version: `1.3.13`
* Scala version: `2.12.10`
* OpenJDK version: `1.8`

nchammas:
Thanks for sharing this report, @Wadaboa. This is the correct place to report issues like this.

I haven't tested Flintrock myself with Spark 3.0 yet, so I'm not surprised there are issues.

> Cannot run program "/usr/local/opt/openjdk@11/bin/java" (in directory "/home/ec2-user/spark/work/driver-20200718133018-0005")

[...]

PR #316, opened 2020-10-21
Add support for configurable JDKs

sfcoy:
* Adds support for user configurable JDKs in order to enable JDK 11 or newer to be installed
* 

I tested this PR by...

1. Creating and running Java 8 based clusters
2. Creating and running Java 11 based clusters

### DIFF
--- a/flintrock/core.py
+++ b/flintrock/core.py
@@ -508,40 +508,42 @@ def get_java_major_version(client: paramiko.client.SSHClient):
             tokens = output.split()
             # First line of the output is like: 'openjdk version "1.8.0_252"' or 'openjdk version "11.0.7" 2020-04-14'
             # Get the version string and strip out the first two parts of the
-            # version as a tuple: (1, 8)
+            # version as an int: 7, 8, 9, 10...
             if len(tokens) >= 3:
                 version_parts = tokens[2].strip('"').split(".")
                 if len(version_parts) >= 2:
                     if version_parts[0] == "1":
                         # Java 6, 7 or 8
-                        return version_parts[1]
+                        return int(version_parts[1])
                     else:
                         # Java 9+
-                        return version_parts[0]
+                        return int(version_parts[0])
         except SSHError:
             pass
 
     return None
 
 
-def ensure_java(client: paramiko.client.SSHClient, jdk: str):
+def ensure_java(client: paramiko.client.SSHClient, java_version: int):
     host = client.get_transport().getpeername()[0]
     java_major_version = get_java_major_version(client)
 
+    if java_major_version and java_major_version == java_version:
+        logger.info("Java {j} is already installed, skipping Java install".format(j=java_major_version))
+    elif java_major_version and java_major_version > java_version:
+        logger.warning("Downgrading existing Java {j} installation".format(j=java_major_version))
+
     # sudo yum install -y java-1.8.0-openjdk
     # sudo amazon-linux-extras install -y java-openjdk11
-    if not java_major_version or java_major_version < jdk:
-        logger.info("[{h}] Installing Java {j}...".format(h=host, j=jdk))
+    elif not java_major_version or java_major_version < java_version:
+        logger.info("[{h}] Installing Java {j}...".format(h=host, j=java_version))
 
         install_adoptopenjdk_repo(client)
-        java_package = "adoptopenjdk-{j}-hotspot".format(j=jdk)
+        java_package = "adoptopenjdk-{j}-hotspot".format(j=java_version)
         ssh_check_output(
             client=client,
             command="""
                 set -e
-
-                # Set up the AdoptOpenJDK yum repo
-                sudo mv /tmp/adoptopenjdk.repo /etc/yum.repos.d/
                 
                 # Install Java first to protect packages that depend on Java from being removed.
                 sudo yum install -q -y {jp}
@@ -559,13 +561,19 @@ def ensure_java(client: paramiko.client.SSHClient, jdk: str):
 
 def install_adoptopenjdk_repo(client):
     """
-    Copies the adoptopenjdk.repo file to the /tmp directory on the client.
-    It must be sudo mv'd to /etc/yum.repos.d/ later.
+    Installs the adoptopenjdk.repo file into /etc/yum.repos.d/
     """
     with client.open_sftp() as sftp:
         sftp.put(
             localpath=os.path.join(SCRIPTS_DIR, 'adoptopenjdk.repo'),
             remotepath='/tmp/adoptopenjdk.repo')
+    ssh_check_output(
+        client=client,
+        command="""
+            # Use sudo to install the repo file
+            sudo mv /tmp/adoptopenjdk.repo /etc/yum.repos.d/        
+        """
+    )
 
 
 def setup_node(
@@ -575,7 +583,7 @@ def setup_node(
         # can be looked up by host and reused?
         ssh_client: paramiko.client.SSHClient,
         services: list,
-        jdk: str,
+        java_version: int,
         cluster: FlintrockCluster):
     """
     Setup a new node.
@@ -617,7 +625,7 @@ def setup_node(
     cluster.storage_dirs.root = storage_dirs['root']
     cluster.storage_dirs.ephemeral = storage_dirs['ephemeral']
 
-    ensure_java(ssh_client, jdk)
+    ensure_java(ssh_client, java_version)
 
     for service in services:
         try:
@@ -635,7 +643,7 @@ def setup_node(
 def provision_cluster(
         *,
         cluster: FlintrockCluster,
-        jdk: str,
+        java_version: int,
         services: list,
         user: str,
         identity_file: str):
@@ -644,7 +652,7 @@ def provision_cluster(
     """
     partial_func = functools.partial(
         provision_node,
-        jdk=jdk,
+        java_version=java_version,
         services=services,
         user=user,
         identity_file=identity_file,
@@ -685,7 +693,7 @@ def provision_cluster(
 
 def provision_node(
         *,
-        jdk: str,
+        java_version: int,
         services: list,
         user: str,
         host: str,
@@ -708,7 +716,7 @@ def provision_node(
         setup_node(
             ssh_client=client,
             services=services,
-            jdk=jdk,
+            java_version=java_version,
             cluster=cluster)
         for service in services:
             service.configure(
@@ -779,7 +787,10 @@ def add_slaves_node(
 
     with client:
         if is_new_host:
-            setup_node(ssh_client=client, services=services, cluster=cluster)
+            setup_node(
+                ssh_client=client,
+                services=services,
+                cluster=cluster)
 
         for service in services:
             service.configure(
--- a/flintrock/core.py
+++ b/flintrock/core.py
@@ -495,7 +495,10 @@ def run_against_hosts(*, partial_func: functools.partial, hosts: list):
             future.result()
 
 
-def get_java_major_version(client: paramiko.client.SSHClient):
+def get_installed_java_version(client: paramiko.client.SSHClient):
+    """
+    :return: the major version (5,6,7,8...) of the currently installed Java or None if not installed
+    """
     possible_cmds = [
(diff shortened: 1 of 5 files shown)


flintrock/core.py
```diff
@@ -506,42 +506,66 @@ def get_java_major_version(client: paramiko.client.SSHClient):
                 client=client,
                 command=command)
             tokens = output.split()
-            # First line of the output is like: 'java version "1.8.0_20"'
+            # First line of the output is like: 'openjdk version "1.8.0_252"' or 'openjdk version "11.0.7" 2020-04-14'
             # Get the version string and strip out the first two parts of the
             # version as a tuple: (1, 8)
```
nchammas, line 511: We should update this comment since we are changing the return type.

flintrock/core.py
```diff
@@ -506,42 +506,66 @@ def get_java_major_version(client: paramiko.client.SSHClient):
...
-def ensure_java8(client: paramiko.client.SSHClient):
+def ensure_java(client: paramiko.client.SSHClient, jdk: str):
     host = client.get_transport().getpeername()[0]
     java_major_version = get_java_major_version(client)
 
-    if not java_major_version or java_major_version < (1, 8):
-        logger.info("[{h}] Installing Java 1.8...".format(h=host))
+    # sudo yum install -y java-1.8.0-openjdk
+    # sudo amazon-linux-extras install -y java-openjdk11
+    if not java_major_version or java_major_version < jdk:
```
nchammas, line 533: What should we do if the host has a newer version of Java than the one the user requested? At the very least, we should perhaps print some kind of warning. Ideally, we should ensure that the requested version of Java is activated even if a newer version is already present.

flintrock/core.py
```diff
@@ -506,42 +506,66 @@ def get_java_major_version(client: paramiko.client.SSHClient):
...
             if len(tokens) >= 3:
                 version_parts = tokens[2].strip('"').split(".")
                 if len(version_parts) >= 2:
-                    return tuple(int(part) for part in version_parts[:2])
+                    if version_parts[0] == "1":
+                        # Java 6, 7 or 8
+                        return version_parts[1]
+                    else:
+                        # Java 9+
+                        return version_parts[0]
```
nchammas, line 520: Do we expect this method's return to always be an integer? If so, I think we should convert these strings to integers accordingly.

Actually, it looks like we are defining the JDK as an integer in the command-line interface, so I think we should match that here.

flintrock/core.py
```diff
@@ -506,42 +506,66 @@ def get_java_major_version(client: paramiko.client.SSHClient):
...
+                # Install Java first to protect packages that depend on Java from being removed.
+                sudo yum install -q -y {jp}
 
                 # Remove any older versions of Java to force the default Java to 1.8.
                 # We don't use /etc/alternatives because it does not seem to update links in /usr/lib/jvm correctly,
                 # and we don't just rely on JAVA_HOME because some programs use java directly in the PATH.
                 sudo yum remove -y java-1.6.0-openjdk java-1.7.0-openjdk
 
-                sudo sh -c "echo export JAVA_HOME=/usr/lib/jvm/jre >> /etc/environment"
+                # sudo alternatives --set java {jp}.x86_64
```
nchammas, line 554: Is this an alternative to modifying `/etc/environment`?

flintrock/core.py
```diff
@@ -506,42 +506,66 @@ def get_java_major_version(client: paramiko.client.SSHClient):
...
+                sudo sh -c "echo export JAVA_HOME=/usr/lib/jvm/{jp} >> /etc/environment"
                 source /etc/environment
-            """)
+            """.format(jp=java_package))
+
+
+def install_adoptopenjdk_repo(client):
+    """
+    Copies the adoptopenjdk.repo file to the /tmp directory on the client.
+    It must be sudo mv'd to /etc/yum.repos.d/ later.
```
nchammas, line 563: Why not do the `mv` as part of this function? Seems like it would be cleaner to have this setup happen in one go.

If you think these steps should be separate, then at the very least I would suggest returning the remote path from here and using that in `ensure_java()`, so that `ensure_java()` doesn't just assume where to find the script.

flintrock/core.py
```diff
@@ -750,10 +779,7 @@ def add_slaves_node(
 
     with client:
         if is_new_host:
-            setup_node(
-                ssh_client=client,
-                services=services,
-                cluster=cluster)
+            setup_node(ssh_client=client, services=services, cluster=cluster)
```
nchammas, line 782: Unintentional whitespace change?

flintrock/flintrock.py
```diff
@@ -253,6 +253,7 @@ def cli(cli_context, config, provider, debug):
 @cli.command()
 @click.argument('cluster-name')
 @click.option('--num-slaves', type=click.IntRange(min=1), required=True)
+@click.option('--jdk', type=click.IntRange(min=8), default=8)
```
nchammas, line 256: I prefer `--jdk-version` or `--java-version` to match the options we have for Spark and Hadoop.

Actually, what's your take on naming this `--jdk-...` vs. `--java-...`? My feeling is that the latter is slightly clearer and more user friendly, but I don't feel strongly about it.

nchammas (commented):
Looks very good to me! Left some minor comments about the details.

One other thought that comes to me is that it would be useful to add Java 8 vs. 11 to the acceptance tests somehow. Not that every test should run twice, once with Java 8 and once with Java 11, but that there should be at least one test that somehow makes sure this Java flag is working for some common use case.

Do you think there is a simple way we can add a test like that?

flintrock/core.py
sfcoy: yup - I have restored this

flintrock/core.py
```diff
@@ -506,42 +507,73 @@ def get_java_major_version(client: paramiko.client.SSHClient):
...
 
 
-def ensure_java8(client: paramiko.client.SSHClient):
+def ensure_java(client: paramiko.client.SSHClient, java_version: int):
     host = client.get_transport().getpeername()[0]
     java_major_version = get_java_major_version(client)
 
-    if not java_major_version or java_major_version < (1, 8):
-        logger.info("[{h}] Installing Java 1.8...".format(h=host))
+    if java_major_version and java_major_version == java_version:
```
nchammas, line 532: I think this can just be `if java_major_version == java_version`.

flintrock/core.py
```diff
@@ -506,42 +507,73 @@ def get_java_major_version(client: paramiko.client.SSHClient):
...
+        logger.info("Java {j} is already installed, skipping Java install".format(j=java_major_version))
+    elif java_major_version and java_major_version > java_version:
+        logger.warning("Downgrading existing Java {j} installation".format(j=java_major_version))
 
+    # sudo yum install -y java-1.8.0-openjdk
+    # sudo amazon-linux-extras install -y java-openjdk11
+    elif not java_major_version or java_major_version < java_version:
+        logger.info("[{h}] Installing Java {j}...".format(h=host, j=java_version))
```
nchammas, line 540: I think here you probably just want to have something like:

```python
else:
    if java_major_version > java_version:
        logger.warning("Downgrading existing Java {j} installation".format(j=java_major_version))
    logger.info("[{h}] Installing Java {j}...".format(h=host, j=java_version))
    ...
```

What you have now will show a warning if the existing Java is newer than what's requested, but it won't actually downgrade the version.

flintrock/flintrock.py
```diff
@@ -722,6 +726,8 @@ def stop(cli_context, cluster_name, ec2_region, ec2_vpc_id, assume_yes):
               help="Path to SSH .pem file for accessing nodes.")
 @click.option('--ec2-user')
 @click.option('--ec2-spot-price', type=float)
+@click.option('--ec2-spot-request-duration', default='7d',
+              help="Duration a spot request is valid (e.g. 3d 2h 1m).")
```
nchammas, line 730: Looks like some recent changes from #315 somehow made it in here.

nchammas (commented):
This looks great! I think a couple final tweaks and it's ready to go. I will also make time to test it myself this week.

By the way, a request: If you can, please avoid force pushing the branch. It disconnects review comments from the PR commit history and makes it a bit harder to follow review comments over time.

I'm guessing you did it to overcome some troubles updating the branch from the latest `master`? It may be easier to just manually delete the spot request duration stuff at this point.

sfcoy (commented):
> One other thought that comes to me is that it would be useful to add Java 8 vs. 11 to the acceptance tests somehow. Not that every test should run twice, once with Java 8 and once with Java 11, but that there should be at least one test that somehow makes sure this Java flag is working for some common use case.
> 
> Do you think there is a simple way we can add a test like that?

It would be nice if we could. However I am not currently up with modern python testing methodology (I'm a Java guy).
Right now I manually test four scenarios:

1. Vanilla AMI with no Java installed
2. Java 11 AMI with java-version: 8
3. Java 11 AMI with java-version: 11
4. Java 11 AMI with java-version: 14

I had to build my own Java 11 AMI because the only existing ones seem to require subscriptions and AWS was being cranky about these today (returning 400 errors).

nchammas:
Just noting for the record that this issue potentially closes #313.

@sfcoy - Sorry about the delay in reviewing the latest changes here. I intend to get to them this upcoming week and get this merged in!

nchammas (approved):
This PR is good to go. If I can trouble you for one more improvement, it would be to display a warning if the user launches a cluster with Hadoop < 3.3 and Java 11. [Only Hadoop 3.3+ are compatible with Java 11+.](https://cwiki.apache.org/confluence/display/HADOOP/Hadoop+Java+Versions)

If you try to launch a cluster with Hadoop 3.2 and Java 11, you get an error like this on starting up HDFS:

```
2021-01-04 00:17:27,413 ERROR [main] namenode.NameNode (NameNode.java:main(1764)) -
 Failed to start namenode.
java.lang.ExceptionInInitializerError
        at org.eclipse.jetty.webapp.WebInfConfiguration.findAndFilterContainerPaths
(WebInfConfiguration.java:185)
...
Caused by: java.lang.IllegalArgumentException: Invalid Java version 11.0.9.1
```

I would probably add the check somewhere here: https://github.com/nchammas/flintrock/blob/22623d4da3ce3fc05a99c93c8077c8064cb2104a/flintrock/flintrock.py#L406-L412

No worries if you can't get to it. In that case, I will push an appropriate fix to this branch before merging it in, if you don't mind.

sfcoy:
> This PR is good to go. If I can trouble you for one more improvement, it would be to display a warning if the user launches a cluster with Hadoop < 3.3 and Java 11. [Only Hadoop 3.3+ are compatible with Java 11+.](https://cwiki.apache.org/confluence/display/HADOOP/Hadoop+Java+Versions)
> 
> If you try to launch a cluster with Hadoop 3.2 and Java 11, you get an error like this on starting up HDFS:
> 
> ```
> 2021-01-04 00:17:27,413 ERROR [main] namenode.NameNode (NameNode.java:main(1764)) -
>  Failed to start namenode.
> java.lang.ExceptionInInitializerError
>         at org.eclipse.jetty.webapp.WebInfConfiguration.findAndFilterContainerPaths
> (WebInfConfiguration.java:185)
> ...
> Caused by: java.lang.IllegalArgumentException: Invalid Java version 11.0.9.1
> ```
> 
> I would probably add the check somewhere here:
> 
> https://github.com/nchammas/flintrock/blob/22623d4da3ce3fc05a99c93c8077c8064cb2104a/flintrock/flintrock.py#L406-L412
> 
> No worries if you can't get to it. In that case, I will push an appropriate fix to this branch before merging it in, if you don't mind.

The Hadoop team must have added a Java version check, because I have actually built and run this combination before.

nchammas:
I've filed #322 to track adding some sort of compatibility check to Flintrock.

Merging this in. Thank you for persisting through the review process @sfcoy and getting this important enhancement in!

sfcoy:
> I've filed #322 to track adding some sort of compatibility check to Flintrock.
> 
> Merging this in. Thank you for persisting through the review process @sfcoy and getting this important enhancement in!

No worries mate. I was actually becoming a little concerned about you, given the Covid situation and all.

PR #335, opened 2021-02-23
feature: add configurable master instance settings

### LINKED ISSUE #82: Slave-only spot instances
Flintrock supports launching clusters using spot instances. However, when doing so, all the cluster instances are launched as spot instances.

Generally, though, you don't want the master to be a spot instance. You want it to stick around even if various slaves are dropped so that your job can still chug along (though at a slower pace) and so that you can re-add slaves to the cluster at a later time.

We should change our support for launching clusters on spot instances such that only the slaves are spot instances.

This work (kinda) depends on #16.

luhhujbb:
* add custom master configuration/slave only spot

Fully tested on our production environment

Fixes #82.

### DIFF
--- a/flintrock/ec2.py
+++ b/flintrock/ec2.py
@@ -844,13 +844,16 @@ def launch(
         key_name,
         identity_file,
         instance_type,
+        master_instance_type,
         region,
         availability_zone,
         ami,
         user,
         security_groups,
         spot_price=None,
+        master_spot_price=None,
         spot_request_duration=None,
+        master_spot_request_duration=None,
         min_root_ebs_size_gb,
         vpc_id,
         subnet_id,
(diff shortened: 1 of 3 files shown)


devyn:
@nchammas can this be merged in some form? it would be nice to have this feature - depending on the application, the master doesn't really need to be very beefy, but having bigger slaves can be better for performance (to a certain extent)

nchammas:
My apologies to both @luhhujbb and @devyn about leaving this PR hanging. I won't have the capacity in the foreseeable future to review and shepherd this PR through to completion.

I would consider running a fork of Flintrock to get this feature. The instructions in the README should help you build your own version. If you have trouble doing that, please reach out and I will try to help.

PR #353, opened 2022-06-20
Feat/save spot log file

emersonfcamara:
* Create a log file with the response os the method request_spot_instances, it makes easier to localize the requests created in the previous request and destroy it before create another spot instance request.

I tested this PR by creating some flintrock clusters, destroying them and checking if the requests instances was successfully terminated.

To use this new strategy using a container you should set one environment variable called LOG_FOLDER so the flintrock will use this variable to know when to create the log file, below are the example using this strategy.

docker run -it --rm \
		-v $$(pwd)/src:/app/mail-ignition/src \
		-v $$(pwd)/chaordic/src:/app/mail-ignition/chaordic/src \
		-v $$(pwd)/chaordic/tools:/app/mail-ignition/chaordic/tools \
		-v $$(pwd)/core/src:/app/mail-ignition/core/src \
		-v $$(pwd)/core/tools:/app/mail-ignition/core/tools \
		-v $$(pwd)/mail-library/src:/app/mail-ignition/mail-library/src \
		-v $$(pwd)/../mail-devops/:/app/mail-devops \
		-e ENVIRONMENT=development \
		**-e LOG_FOLDER=/app/mail-ignition/core/tools/flintrock/logs \**
		retargeting/mail-ignition /bin/bash

To use the same strategy in a linux server just set the environment variable using export command like below.

**export LOG_FOLDER=/app/mail-ignition/core/tools/flintrock/logs**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ install:
 script:
   - "py.test ./tests/test_static.py"
   - "py.test ./tests/test_flintrock.py"
+  - "py.test ./tests/test_core.py"
   - "pip install -r requirements/maintainer.pip"
   - "py.test ./tests/test_pyinstaller_packaging.py"
 addons:
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,3 +16,5 @@ addons:
     paths:
       - '$(ls -1 dist/*.zip | tr "\n" ":")'
   s3_region: "us-east-1"
+notifications:
+  email: false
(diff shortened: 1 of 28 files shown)


nchammas:
@emersonfcamara - I think #362 should address any problems with Flintrock leaving behind malformed nodes from failed launches, including spot instances. If that PR doesn't address your problem, please chime in and I'd be happy to revisit this issue.

I also see that this PR includes some unrelated changes, so if you want me to take another look at this work then it'll probably be best if you create a new PR that contains just your changes.

Issue #346, opened 2021-10-15
SPARK_PUBLIC_DNS is incorrectly set when launching into a private VPC


maxpoulain:
Hi,

We are having issues in the Spark UI notably when using Flintrock. 
To have more context on our use of Flintrock:

- Flintrock version: 2.0.0
- Python version: 3.8.9
- OS: ubuntu:20.04
- We are using Flintrock in a private VPC with the option `authorize-access-from` in the config file.
- We download Spark and Hdfs sources from our own S3 bucket.

When we go to the Spark UI on port 8080, the page is displayed correctly but the links to the other pages are broken. 
Here you can find an extract from the HTML code of the page for a link to a worker.
```
<a href="http://<?xml version=&quot;1.0&quot; encoding=&quot;iso-8859-1&quot;?>
<!DOCTYPE html PUBLIC &quot;-//W3C//DTD XHTML 1.0 Transitional//EN&quot;
		 &quot;http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd&quot;>
<html xmlns=&quot;http://www.w3.org/1999/xhtml&quot; xml:lang=&quot;en&quot; lang=&quot;en&quot;>
 <head>
  <title>404 - Not Found</title>
 </head>
 <body>
  <h1>404 - Not Found</h1>
 </body>
</html>:8081">
              worker-20211015082848-<MASKED_IP>-42451
            </a>
```
I have a similar error message when I launch `spark-shell` for example:
```
Spark context Web UI available at http://<?xml version="1.0" encoding="iso-8859-1"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
		 "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
  <title>404 - Not Found</title>
 </head>
 <body>
  <h1>404 - Not Found</h1>
 </body>
</html>:4040
```
I have the impression that the error comes from a problem finding the ip address or something related.. maybe we have done a mistake with our configuration or maybe it's not related to Flintrock.. 

So, if you have any clues or elements to solve this problem, this would be a great help for us.

Thank you in advance for your help,

Maxime

maxpoulain:
It seems that `http://169.254.169.254` is not working anymore no ?

nchammas:
[...]

PR #358, opened 2023-03-31
Amazon linux 2023 fix

pragnesh:
* added **ebs-volume-type** config which allow ebs volume type from configuration since amazon linux 2023 by default comes with gp3 volume image
* updated **setup-ephemeral-storage.py** since new version of lsblk doesn't produce same output for setup.
* moved python package installation before running setup-ephemeral-storage.py
* instance data IMDSv2 fix for spark public hostname

I tested this PR by...
launching spark cluster with latest amazon linux 2023 and amazon linux 2 image. both working fine

### DIFF
--- a/flintrock/core.py
+++ b/flintrock/core.py
@@ -632,6 +632,17 @@ def setup_node(
     Cluster methods like provision_node() and add_slaves_node() should
     delegate the main work of setting up new nodes to this function.
     """
+    # TODO: Move Python and Java setup to new service under services.py.
+    #       New service to cover Python/Scala/Java: LanguageRuntimes (name?)
+    ssh_check_output(
+        client=ssh_client,
+        command=(
+            """
+            set -e
+            sudo yum install -y python3 python
+            """
+        )
+    )
     host = ssh_client.get_transport().getpeername()[0]
     ssh_check_output(
         client=ssh_client,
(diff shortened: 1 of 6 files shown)


nchammas:
Did you test this PR against a private VPC btw? I'm wondering if #346 is affected at all by these changes.

pragnesh:
Hi @nchammas ,

setup-ephemeral-storage.py is not running without python package, 

amazon linux 2 has default python installed so it was working earlier but amazon linux 2023 need python to be installed so moved it before setting running setup-ephemeral-storage.py

instance data [IMDSv2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html) is enabed by default for amazon linux 2023, so SPARK_PUBLIC_DNS need to be adjusted for functional spark web UI

pragnesh:
> Did you test this PR against a private VPC btw? I'm wondering if #346 is affected at all by these changes.

no i have not.

pragnesh:
> Did you test this PR against a private VPC btw? I'm wondering if #346 is affected at all by these changes.

I have updated PR to correctly set SPARK_PUBLIC_DNS in private VPC.
I have tested change on amazon linux 2 ami.
I have tested amazon linux 2023 ami with java preinstalled since Adoptium does not work with amazon linux 2023.
I have tried testing private vpc but due to some vpc network configuration i am unable to launch but SPARK_PUBLIC_DNS is set correctly within private vpc.

nchammas:
Sorry to leave this PR hanging. It is still on my mind and I will hopefully pick it up after I get Flintrock 2.1 out.

PR #366, opened 2023-11-20
Refactor how we launch spot instances

### LINKED ISSUE #364: Spot instance tags issue
* Flintrock version: Master version
* Python version: 3.8.10
* OS: Mac M1

Hi !

It seems that there is an issue regarding `TagSpecifications` in `Make finding cluster nodes more reliable in the face of launch failures`. As we are interested in this new feature, we tried to make a release in our forked version but we got this issue:
```
Parameter validation failed:
Unknown parameter in LaunchSpecification: "TagSpecifications", must be one of: SecurityGroupIds, SecurityGroups, AddressingType, BlockDeviceMappings, EbsOptimized, IamInstanceProfile, ImageId, InstanceType, KernelId, KeyName, Monitoring, NetworkInterfaces, Placement, RamdiskId, SubnetId, UserData
```

If we are not wrong aout this issue, it seems that this issue comes from `TagSpecifications` in `common_launch_specs` that is a parameter of `request_spot_instances` and not a key inside `LaunchSpecification` dict as you can see [here](https://boto3.amazonaws.com/v1/documentation/api/1.21.44/reference/services/ec2.html?highlight=create_instance#EC2.Client.request_spot_instances). Also, we can't put `ResourceType': 'instance'` in `client.request_spot_instances` as explained in the [documentation](https://boto3.amazonaws.com/v1/documentation/api/1.21.44/reference/services/ec2.html?highlight=create_instance#EC2.Client.request_spot_instances):
```
TagSpecifications (list) --
The key-value pair for tagging the Spot Instance request on creation. The value for ResourceType must be spot-instances-request , otherwise the Spot Instance request fails. To tag the Spot Instance request after it has been created, see CreateTags.
```
So, I think, we need to put `ResourceType': 'spot-instances-request'` if we want to tag the spot request instance (not spot instance), if not we can remove `TagSpecifications` from spot request instance. 

Here is a possible solution in `_create_instances`:
1. Remove `TagSpecifications` from `common_launch_specs`
2. Add `TagSpecifications=[{'ResourceType': 'spot-instances-request', 'Tags': tag_specifications}],` in `client.request_spot_instances`  
3. Add `TagSpecifications=[{'ResourceType': 'instance', 'Tags': tag_specifications}],` in `ec2.create_instances`
4. Add tags to spot instances using `create_tags` as last step inside `if spot_price:`:
```python
(ec2.instances
             .filter(
                Filters=[
                    {'Name': 'instance-id', 'Values': [instance.id for instance in cluster_instances]}
                ])
             .create_tags(Tags=tag_specifications))
```

We have done these modifications in [our fork](https://github.com/adotmob/flintrock/commit/1533553efeb7a23aa7c565d13a9225802fc472bc), so if you're okay with that, we can do the same here or if you have maybe a more elegant way to do that it could be interesting. 

I hope it's clear and this helps !

Maxime

nchammas:
This PR refactors how we launch spot instances to migrate away from [RequestSpotInstances, which is deprecated][1].

Instead, we now use the same method -- [RunInstances][2] -- whether we are launching on-demand or spot instances. This simplifies the launch code.

This PR also deprecates the `--ec2-spot-request-duration` option, which does not apply to one-time spot instances (as opposed to persistent spot instances). The distinction between one-time vs. persistent did not exist in the RequestSpotInstances API. However, one-time spot instances map most closely to Flintrock's existing handling of spot instances.

In the future, as [previously mentioned here][3], it would be good to somehow delegate Flintrock's handling of all the various EC2 options down to boto3, instead of exposing individual options and having to track changes to the EC2 API. That wouldn't have spared us this migration from RequestSpotInstances to RunInstances, but it would let users get more of what EC2 can do out of the box without requiring Flintrock to explicitly enable it. But that's a future PR.

I tested this PR by running the full test suite, including acceptance tests, as well as manual tests of launching spot instance clusters and add spot nodes to existing  clusters.

Related to #315.
Fixes #364.

[1]: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_RequestSpotInstances.html
[2]: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_RunInstances.html
[3]: https://github.com/nchammas/flintrock/pull/315#pullrequestreview-511265668


nchammas:
cc @mblackgeo - If you're still using Flintrock, I'm curious to know if you agree with my summary of why we don't need `--ec2-spot-request-duration` anymore. (Asking since you added this feature in #315.)

nchammas:
cc @maxpoulain

mblackgeo:
> cc @mblackgeo - If you're still using Flintrock, I'm curious to know if you agree with my summary of why we don't need `--ec2-spot-request-duration` anymore. (Asking since you added this feature in #315.)

Hey thanks for looping me in! Yes we indeed still use Flintrock and with this change to one-time requests it makes sense that the spot request duration is no longer needed. I assume that a `Ctrl + C` when launching instances with Flintrock will cancel the boto/ec2 request?

nchammas:
That is correct, Flintrock will cancel the request and terminate any instances that may have launched as part of the same operation. If you find a case where that doesn't happen, that's a bug.

maxpoulain:
Hi ! 

Thanks for the cc ! 

I just have a question and I haven't found any information in boto documentation.. 

As `ec2.create_instances` is going to be used for spot instances creation and `ec2.create_instances` returns a list of instances. What's happen if there is not enough available instances for example ? Is `ec2.create_instances` going to return provided instances even if it's not the number of instances wanted ? or `ec2.create_instances` is going to wait until all instances are available ? or `ec2.create_instances` is just going to fail ?

maxpoulain:
Yes, it meets our expectations perfectly, it's better/cleaner than using RequestSpotInstances !

PR #373, opened 2023-12-20
Upgrade to paramiko 3.4.0

alex:
This resolves https://terrapin-attack.com/


nchammas:
@alex - Thank you for taking the time to author this upgrade, and sorry for taking so long to merge it.

I modified your PR by re-pinning the dependencies on Python 3.8, as I am not sure whether using Python 3.11 to pin dependencies will select versions that also work on Python 3.8. This point was not clear in the contributing guide, so I also tweaked that as part of this PR.

alex:
No problem. Happy to help.

PR #378, opened 2024-07-05
Fixed centralize logic TODOs in ec2

tomahawk360:
This PR makes the following changes:
* Replaced three security_groups.filter() instances at lines 180, 197 and 379, each marked with a "Centralize logic..." TODO comment, with get_security_groups() function.
* Removed "Centralize logic..." TODO comments in all previously mentioned instances.

I tested this PR by:
* Launching a cluster
* Adding slaves
* Removing slaves
* Destroying a cluster
* Running "pytest" command

Fixes #377.

PR #381, opened 2024-12-11
fix: handle destroy with missing security group

karsten-adot:
* Destroying cluser even if security group is missing (for example in case of ungraceful stop during a previous destroy)

I tested this PR by hand and against incomplete cluster (missing security group)


nchammas:
Hi @karsten-adot and thank you for this fix. I took the liberty of simplifying it a bit given the recent changes we made in #378.

Overall, the code to manage security groups could use a more significant refactoring, but this fix is still helpful.

PR #380, opened 2024-10-12
chore: bump pyyaml to build with py3.13

chenrui333:
This PR makes the following changes:
*  bump pyyaml to build with py3.13

relates to https://github.com/Homebrew/homebrew-core/pull/194055

PR #383, opened 2024-12-13
Drop support for Python 3.8, add support for Python 3.13

nchammas:
* Drop support for Python 3.8, which is [EOL](https://devguide.python.org/versions/).
* Add CI build for Python 3.13.
* Refresh dependencies.

I tested this PR by launching and destroying a cluster.

Fixes #380.


nchammas:
@chenrui333 The 3.13 build passes without bumping PyYAML (I'm looking at the build status for [f7963df](https://github.com/nchammas/flintrock/pull/383/commits/f7963df6149272e375f4be87f2606f6dcf12ddd0)), but based on #380 I looked into it and do see [this changelog entry][1], so I've bumped PyYAML anyway.

[1]: https://github.com/yaml/pyyaml/blob/69c141adcf805c5ebdc9ba519927642ee5c7f639/CHANGES#L12

